PR #27, opened 2022-11-17
Add release workflows

gaiksaya:
### Description
This PR adds below things:
* A GitHub action workflow that will be triggered when a tag is pushed to this repository. The workflow builds the product and creates a draft release with built artifacts attached as artifacts.tar.gz in the release.
* A jenkins workflow that is triggered based on above draft release. The workflow looks for `artifacts.tar.gz`, downloads it and then signs it and publishes it to [PyPi](https://pypi.org/) 
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,31 @@
+name: Release drafter
+
+on:
+  push:
+    tags:
+      - "*"
```
dhrubo-os, line 6: Can we add a comment here: `# Push events to every tag not containing /`

gaiksaya: Sorry I didn't quite get that. This workflow will be triggered irrespective of any type of tag. Is there a plan to push specifically named tags?
Also my understanding is tag will be pushed when you want to release this product. Correct me if I am wrong please.

dhrubo-os: From the [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags) it looks like `*` is for `Matches all branch and tag names that don't contain a slash (/).`

I just wanted to add that here as a comment.

gaiksaya: The `tags` options on line 5 will look only for tags and not branches. I can add a comment you asked just for as an information.

dhrubo-os: Yes please.

dhrubo-os (approved):

gaiksaya:
Making some changes on core library. We found out that the api tokens used to publish the packages on pypi can be scoped to per package. We are working on making some changes. Please wait till I make some changes in this PR.
Thanks!

dhrubo-os (approved):

.github/workflows/release-drafter.yml
gaiksaya: Done

dhrubo-os (approved):

mingshl (approved):

PR #52, opened 2022-12-08
Unit test for SentenceTransformerModel

mingshl:
### Description
Unit test for SentenceTransformerModel and point torch version less than 1.13.0 to try fixing the security test [CVE-2022-45907](https://www.mend.io/vulnerability-database/CVE-2022-45907)
 
### Issues Resolved
Unit test for SentenceTransformerModel and point torch version less than 1.13.0 to try fixing the security test
 
### Check List
  - [ ] All tests pass


b4sjoo (approved):

opensearch_py_ml/sentence_transformer_model/sentencetransformermodel.py
```diff
@@ -917,9 +918,13 @@ def make_model_config_json(
...
     def __qryrem(self, x):
-        # for removing the "QRY:" token
-        y = x.passages.split(" QRY:")[0].split("<|startoftext|>")
-        return y[1]
+        # for removing the "QRY:" token if they exist in passages
+        if "QRY:" in x.passages and "<|startoftext|>" in x.passages:
+            y = x.passages.split(" QRY:")[0].split("<|startoftext|>")
+            return y[1]
+        else:
+            print("QRY: and <|startoftext|> not found in passages. ")
```
ylwu-amzn, line 926: Why we need to remove "QRY:" and "<|startoftext|>" ?

Message should be ?
```suggestion
            print("QRY: or <|startoftext|> not found in passages. ")
```

Is it possible that this line will print out too many messages if thousands lines doesn't have "QRY"?

requirements-dev.txt
```diff
@@ -5,7 +5,7 @@ pandas>=1.5,<2
 matplotlib>=3.6.0,<4
 numpy>=1.23.3,<2
 opensearch-py>=2
-torch>=1.13.0
+torch<1.13.0
```
ylwu-amzn, line 8: Why we need a lower version?

ylwu-amzn: > In PyTorch before trunk/89695, torch.jit.annotations.parse_type_line can cause arbitrary code execution because eval is used unsafely.

What does it mean "before trunk/89695"?

mingshl: probably it's referring the torch repo before this merge https://github.com/pytorch/pytorch/commit/767f6aa49fe20a2766b9843d01e3b7f7793df6a3. after this merge, it trigger an unsafe issue in the new version and torch team is still working on it. 

It's new issue that happened after 11/26/2022 lately https://vuldb.com/?id.214417

mingshl: so the best thing we can do is avoid this new version until the issue is resolved by Pytorch team

ylwu-amzn (approved):

requirements-dev.txt
dhrubo-os: If we want to avoid this version let's put: `torch=1.12.0`.

PR #63, opened 2023-01-03
 bug fix of sentencetransformermodel + add integration test from mode…

dhrubo-os:
Signed-off-by: Dhrubo Saha <dhrubo@amazon.com>

### Description
[bug fix of sentencetransformermodel + add integration test from mode training to loading model in opensearch cluster
 ]
 
### Issues Resolved
[57, 58, 59, 60]
 
### Check List
  - [ ] All tests pass


tests/ml_models/test_sentencetransformermodel_pytest.py
```diff
@@ -58,6 +58,8 @@ def test_check_attribute():
     assert test_model.model_id == "sentence-transformers/msmarco-distilbert-base-tas-b"
 
     default_folder = os.path.join(os.getcwd(), "model_files")
+    print(default_folder)
```
mingshl, line 61: is it better to remove the print message in this test file?

mingshl (commented):
There are three files added in the docs, is it on purpose?  synthetic_queries_batch_0, synthetic_queries_batch_1, synthetic_queries_batch_2. 

the class read files from zip file, those pickled files are not serving as direct reading files. 

[docs/source/examples/synthetic_queries/synthetic_queries_batch_0.p](https://github.com/opensearch-project/opensearch-py-ml/pull/63/files#diff-26ea3437e39830cd5fab16abdf1aecf8af05c5712dee32b88f038e2e48a04fca)

tests/ml_models/test_sentencetransformermodel_pytest.py
dhrubo-os: Good catch, will remove.

And about the `three files added in the docs` question: these files were added by the sphinx due to converting the notebook to html. We didn't add any files on purpose.

docs/source/examples/demo_transformer_model_train_save_upload_to_openSearch.ipynb
```diff
@@ -163,19 +167,18 @@
...
-      "100%|██████████| 243/243 [00:00<00:00, 629923.28it/s]"
+      "100%|██████████| 243/243 [00:00<00:00, 148986.39it/s]"
      ]
     },
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "generated config file: at /home/ec2-user/.cache/huggingface/accelerate/default_config.yaml\n",
+      "generated config file: at /Users/dhrubo/.cache/huggingface/accelerate/default_config.yaml\n",
```
ylwu-amzn, line 177: Let's avoid using personal home folder

opensearch_py_ml/ml_commons/ml_common_client.py
```diff
@@ -77,3 +78,56 @@ def load_model(self, model_id: str):  # type: ignore
...
+
+        MODEL_TASK_API_ENDPOINT = f"tasks/{task_id}"
+        API_URL = f"{ML_BASE_URI}/{MODEL_TASK_API_ENDPOINT}"
+
+        return self._client.transport.perform_request(
+            method="GET",
+            url=API_URL,
+        )
+
+    def generate_embedding(self, model_id: str, sentences: List[str]):  # type: ignore
```
ylwu-amzn, line 106: Not for this line. 
How about adding methods for all other REST APIs like `unload`, `delete model`, `get model` etc.?

tests/ml_models/test_sentencetransformermodel_pytest.py
mingshl: weird, previously the notebook read the zipfile of synthetic queries in the same way and it didn't add the unzipped .p files to the doc folder. The only concern is that I am not sure if we have permission to post the data file on the github. I believed I generated the file using the dataset package data.

docs/source/examples/demo_transformer_model_train_save_upload_to_openSearch.ipynb
dhrubo-os: I think these information are too much to show. I will add `isVerbose` field in this method to control the visibility of these messages in the next PR.

opensearch_py_ml/ml_commons/ml_common_client.py
dhrubo-os: Good idea. Can I add these in a separate PR?

tests/ml_models/test_sentencetransformermodel_pytest.py
dhrubo-os: This is a binary file. I already uploaded a same synthetic query zip file for the integration test. I don't think that should be an issue. 

Also I'm planning to raise another PR to support other rest API. Can I remove the print message in that time?

mingshl: cool

opensearch_py_ml/ml_commons/ml_common_client.py
ylwu-amzn: Sure, that's ok

dhrubo-os: Changed my mind, added all these 3 with integration test and documentation as well.

opensearch_py_ml/ml_models/sentencetransformermodel.py
```diff
@@ -163,12 +188,30 @@ def train(
...
                 )
             else:
-                subprocess.run(
-                    [
-                        "accelerate launch self.train_model(train_examples, self.model_id, output_model_path, output_model_name,use_accelerate,learning_rate,num_epochs,verbose)"
-                    ]
-                )
-        else:
+                try:
+                    subprocess.Popen(
```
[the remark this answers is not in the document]
dhrubo-os: > I am curious why changing `subprocess.run` to `subprocess.Popen`? Normally, 'The recommended approach to invoking subprocesses is to use the [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) function for all use cases it can handle.'

[subprocess.run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) [was added in Python 3.5](https://docs.python.org/3/whatsnew/3.5.html#subprocess) as a simplification over [subprocess.Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) when you just want to execute a command and wait until it finishes, but you don't want to do anything else in the mean time. For other cases, you still need to use subprocess.Popen.

The main difference is that subprocess.run() executes a command and waits for it to finish, while with subprocess.Popen you can continue doing your stuff while the process finishes and then just repeatedly call [Popen.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) yourself to pass and receive data to your process. Secondly, subprocess.run() returns [subprocess.CompletedProcess](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess).

You could say with `Popen` you can achieve some sort of asynchronous programming which I tried to do to generate some output. But in the end it didn't improve the code so I reverted now to the simplest approach.

mingshl (approved):

opensearch_py_ml/ml_commons/ml_common_client.py
```diff
@@ -77,3 +78,135 @@ def load_model(self, model_id: str):  # type: ignore
             method="POST",
             url=API_URL,
         )
+
+    def get_task_info(self, task_id: str):  # type: ignore
```
ylwu-amzn, line 82: How about just name the method as `get_task`?
Same comment for `get_model_info`, change to `get_model`?

ylwu-amzn: Can we remove `# type: ignore` ?

opensearch_py_ml/ml_commons/ml_common_client.py
```diff
@@ -77,3 +78,135 @@ def load_model(self, model_id: str):  # type: ignore
...
+        :type task_id: string
+
+        Returns
+        -------
+        :return: returns a json object, with detailed information about the task
+        :rtype: object
+        """
+
+        MODEL_TASK_API_ENDPOINT = f"tasks/{task_id}"
+        API_URL = f"{ML_BASE_URI}/{MODEL_TASK_API_ENDPOINT}"
```
ylwu-amzn, line 99: Seems combine these two lines to one line is more readable.
```suggestion
        API_URL = f"{ML_BASE_URI}/tasks/{task_id}"
```

opensearch_py_ml/ml_commons/ml_common_client.py
```diff
@@ -5,6 +5,9 @@
 # Any modifications Copyright OpenSearch Contributors. See
 # GitHub history for details.
 
+
+from typing import List
+
 from opensearchpy import OpenSearch
 
 from opensearch_py_ml.ml_commons.ml_common_utils import ML_BASE_URI
```
ylwu-amzn, line 13: I'm not sure for python, but for Java, we generally put util method (usually static method) in util class. But I see we just keep several constants in https://github.com/opensearch-project/opensearch-py-ml/blob/main/opensearch_py_ml/ml_commons/ml_common_utils.py, that seems not necessary to keep this file. How about just move these constants to this `ml_common_client.py` to keep the tightly coupled code closer?

opensearch_py_ml/ml_commons/ml_common_client.py
```diff
@@ -5,6 +5,9 @@
 # Any modifications Copyright OpenSearch Contributors. See
 # GitHub history for details.
 
+
```
ylwu-amzn, line 8: Not for this line.
The file name seems mismatch with ml-commons plugin name https://github.com/opensearch-project/ml-commons

How about change to `ml_commons_client` ?

opensearch_py_ml/ml_commons/ml_common_client.py
dhrubo-os: Yeah idea is also similar here to keep common method and variables in the util class which can be reused in multiple files. These variables are used also in other files like model_uploader.py

ylwu-amzn (approved):
LGTM. You can address comments in a new PR.

Issue #104, opened 2023-03-09
[FEATURE] MLCommons Integration


dhrubo-os:
**Is your feature request related to a problem?**

We want to integrate all the ML-Commons api to opensearch-py-ml.

Right now following apis are supported:

1. Upload Model
2. Load model
3. Get task information
4. Get model information
5. Unload model
6. Delete model

We want to have support for following apis:

1. Train --> https://opensearch.org/docs/latest/ml-commons-plugin/api/#train-model
2. Predict --> https://opensearch.org/docs/latest/ml-commons-plugin/api/#predict
3. Train and Predict  --> https://opensearch.org/docs/latest/ml-commons-plugin/api/#train-and-predict
4. Search Task --> https://opensearch.org/docs/latest/ml-commons-plugin/api/#search-task
5. Delete Task --> https://opensearch.org/docs/latest/ml-commons-plugin/api/#delete-task                   ---------> This is done.
6. Profile api --> https://opensearch.org/docs/latest/ml-commons-plugin/api/#profile
7. Search model --> https://opensearch.org/docs/latest/ml-commons-plugin/api/#search-model
8. Execute api --> https://opensearch.org/docs/latest/ml-commons-plugin/api/#execute : For testing after development, we can use this request:

```
POST _plugins/_ml/_execute/local_sample_calculator
{
    "operation": "max",
    "input_data": [1.0, 2.0, 3.0]
}
```
So, to implement this request python function can expect two variables:
1. algorithm name
2. input request --> here we need to check the input validation like if the input string is a valid json object or not.

Merey1508:
Hi, everyone! I wanted to let you know that I am currently focusing my efforts on the Profile API (№6). 

@dhrubo-os, I've noticed that we have two identical tasks (№4 and №7) in our API support list. Perhaps we can review and clarify if there are any differences between the two, or if one of them can be removed to avoid confusion?

dhrubo-os:
> 

Thanks.

dhrubo-os:
> Hi, everyone! I wanted to let you know that I am currently focusing my efforts on the Profile API (№6).
> 
> @dhrubo-os, I've noticed that we have two identical tasks (№4 and №7) in our API support list. Perhaps we can review and clarify if there are any differences between the two, or if one of them can be removed to avoid confusion?

@Merey1508 Yes, you are right. Removed the duplicate task. Thanks

dhrubo-os:
[...]

PR #110, opened 2023-03-19
Added search model API

Nurlanprog:
### Description
Search of models can be done in 2 ways: 1) Search models by default without parameters 2) Search all models with given algorithm name. I tested with and without the parameter. 
Finally, I am not added "size" parameter (models number) but I am tested it. If I will specify "size" approximately less than 800 it works, but if I will specify 1000 then It will throw error: read time out. Let me know if "size" parameter is needed.
 
### Issues Resolved
Link to issue: https://github.com/opensearch-project/opensearch-py-ml/issues/104
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/opensearch_py_ml/ml_commons/ml_commons_client.py
+++ b/opensearch_py_ml/ml_commons/ml_commons_client.py
@@ -198,3 +198,21 @@ class MLCommonClient:
             method="DELETE",
             url=API_URL,
         )
+    
+    def search_model(self, algorithm_name: str = "") -> object:
+        API_URL = f"{ML_BASE_URI}/models/_search"
+
+        if len(algorithm_name) == 0:
+            API_BODY = {"query": {"match_all": {}}}
+        else:
+            API_BODY = {
+                "query": {
+                    "term": {"algorithm": {"value": algorithm_name}}
+                }
+            }
+
+        return self._client.transport.perform_request(
+            method="POST",
+            url=API_URL,
+            body=API_BODY,
+        )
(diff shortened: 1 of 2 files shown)


opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -198,3 +198,21 @@ def delete_model(self, model_id: str) -> object:
             method="DELETE",
             url=API_URL,
         )
+    
+    def search_model(self, algorithm_name: str = "") -> object:
```
dhrubo-os, line 202: Thanks for raising the PR.

In search we can search multiple ways. I understand you were covering both ways from [here](https://opensearch.org/docs/2.6/ml-commons-plugin/api/#search-model). But these are just two difference example.

Let's think about how else customer can provide search query.

For example :

```
POST /_plugins/_ml/models/_search
{
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "algorithm": "KMEANS"
          }
        }
      ]
    }
  }
}
```

I would like to cover all the possibilities we can search for model in this function.

tests/ml_commons/test_ml_commons_client.py
```diff
@@ -165,3 +165,21 @@ def test_integration_model_train_upload_full_cycle():
                 except:  # noqa: E722
                     raised = True
                 assert raised == False, "Raised Exception in deleting model"
+
+                try:
+                    search_model_obj = ml_client.search_model()
```
dhrubo-os, line 170: We don't have to add test here. We can add separate test method for search api testing. This method was mostly covering the full cycle of the model integration.

In the search api we should check how many models we got from the search result. For example if we think there should be one "TEXT_EMBEDDING" model, we should count the output of the search result and assert if the number matches or not.

Nurlanprog: Ok, thank you! I will fix it.

PR #111, opened 2023-03-19
Upload pretrained models

### LINKED ISSUE #106: [FEATURE] add support for MLCommons Pre-trained models support

No.

From Opensearch 2.6 version, MLCommons support [Pre-trained models](https://opensearch.org/docs/latest/ml-commons-plugin/pretrained-models/).

But from opensearch-py-ml's [upload_model](https://github.com/opensearch-project/opensearch-py-ml/blob/main/opensearch_py_ml/ml_commons/ml_commons_client.py#L28-L33) doesn't support that yet. 

Either we need to add another function `upload_pretrained_model` or include uploading pre-trained model feature in `upload_model` method

AlibiZhenis:
### Description
Added a feature of uploading pretrained models. Closes #106 .
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,7 +12,7 @@ jobs:
         cluster: ["opensearch"]
         secured: ["true"]
         entry:
-          - { opensearch_version: 2.5.0 }
+          - { opensearch_version: 2.6.0 }
 
     steps:
       - name: Checkout
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,7 +12,7 @@ jobs:
         cluster: ["opensearch"]
         secured: ["true"]
         entry:
-          - { opensearch_version: 2.6.0 }
+          - { opensearch_version: 2.5.0 }
 
     steps:
       - name: Checkout
--- a/opensearch_py_ml/ml_commons/ml_commons_client.py
+++ b/opensearch_py_ml/ml_commons/ml_commons_client.py
@@ -97,15 +97,20 @@ class MLCommonClient:
 
         # loading the model chunks from model index
         if load_model:
-            task = self.load_model(model_id)
-            output = self.get_task_info(task.get("task_id"), wait_until_task_done=True)
+            self.load_model(model_id)
+            for i in range(120):  # timeout is 120 seconds
+                time.sleep(1)
+                ml_model_status = self.get_model_info(model_id)
+                model_state = ml_model_status.get("model_state")
+                if model_state != "LOADING":
+                    break
 
-            if output["state"] == "COMPLETED":
+            if model_state == "LOADED":
                 print("Model loaded into memory successfully")
-            elif output["state"] == "COMPLETED_WITH_ERROR":
-                print("Model loaded into memory with error")
+            elif model_state == "PARTIALLY_LOADED":
+                print("Model was loaded into memory only partially")
             else:
-                raise Exception(output["error"])
+                raise Exception("Model load failed")
 
         return model_id
 
--- a/opensearch_py_ml/ml_commons/ml_commons_client.py
+++ b/opensearch_py_ml/ml_commons/ml_commons_client.py
@@ -64,29 +64,33 @@ class MLCommonClient:
         )
 
     def upload_pretrained_model(
-        self, name: str, version: str, model_format: str, load_model: bool = True
+        self,
+        model_name: str,
+        model_version: str,
+        model_format: str,
+        load_model: bool = True,
     ):
         """
         This method uploads a pretrained model into opensearch cluster using ml-common plugin's api.
         First, this method creates a model id to store model info
-        and then loads the model from Hugging Face if load_model is True.
+        and then loads the model in memory if load_model is True.
         The model has to be supported by ML Commons. Refer to https://opensearch.org/docs/latest/ml-commons-plugin/pretrained-models/.
 
-        :param name: Name of the pretrained model
-        :type name: string
-        :param version: Version of the pretrained model
-        :type version: string
-        :param model_format: Currently only supports "TORCH_SCRIPT"
+        :param model_name: Name of the pretrained model
+        :type model_name: string
+        :param model_version: Version of the pretrained model
+        :type model_version: string
+        :param model_format: "TORCH_SCRIPT" or "ONNX"
         :type model_format: string
-        :param load_model: Whether to load the model chunks from Hugging Face
+        :param load_model: Whether to load the model in memory using uploaded model chunks
         :type load_model: bool
-        :return: returns the model_id so that we can use this for further operation.
+        :return: returns the model_id so that we can use this for further operation
         :rtype: string
         """
         # creating model meta doc
         model_config_json = {
-            "name": name,
-            "version": version,
+            "name": model_name,
+            "version": model_version,
             "model_format": model_format,
         }
         model_id = self._send_model_info(model_config_json)
@@ -110,9 +114,9 @@ class MLCommonClient:
         This method sends the pretrained model info to ML Commons' upload api
 
         :param model_meta_json: a dictionary object with model configurations
-        :type model_meta_json: string
+        :type model_meta_json: dict
         :return: returns a unique id of the model
-        :rtype: int
+        :rtype: string
         """
         output: Union[bool, Any] = self._client.transport.perform_request(
             method="POST",
--- a/opensearch_py_ml/ml_commons/ml_commons_client.py
+++ b/opensearch_py_ml/ml_commons/ml_commons_client.py
@@ -63,10 +63,11 @@ class MLCommonClient:
             model_path, model_config_path, isVerbose
         )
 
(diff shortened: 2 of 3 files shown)


opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -63,6 +63,71 @@ def upload_model(
             model_path, model_config_path, isVerbose
         )
 
+    def upload_pretrained_model(self, name: str, version: str, model_format):
```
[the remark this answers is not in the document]
AlibiZhenis: So, if I'm not mistaken, to upload a pretrained model, we have to:

1. Send model info to /models/_upload
2. Then load the model using load model api, after which ML Commons downloads the model chunks from Hugging Face by itself.

If I do the above 2 steps, I can successfully use the model for inference in dev tools.

Are you saying that that I shouldn't do the second part? Because without it, if I try to do inference, I get the "model not loaded" error. 

I would appreciate it if you clarified this for me.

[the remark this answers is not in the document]
AlibiZhenis: I think it's a great idea. The argument was added.

Also, about sending model info to upload api, that's what _create_model_metadoc method does anyway. I just misnamed the method.

Both of these issues were fixed in the last commit.

dhrubo-os:
Thanks for the PR. It looks good. 

Integration test is failing. I believe this is because right now our integration test is based on [opensearch 2.5.0](https://github.com/opensearch-project/opensearch-py-ml/blob/main/.github/workflows/integration.yml#L15)

But this new feature is in `opensearch 2.6.0`. So we should update that version to pass the integration test.

Lint check is also failing. [Check here](https://github.com/opensearch-project/opensearch-py-ml/blob/main/DEVELOPER_GUIDE.md#to-test-code-formatting-and-linting-issue-we-can-run)

AlibiZhenis:
So what should I do next? I have opensearch-py 2.2.0 installed, which pip says is the latest version.

dhrubo-os:
> So what should I do next? I have opensearch-py 2.2.0 installed, which pip says is the latest version.



> 

Can you please change to `2.6.0` [here])https://github.com/opensearch-project/opensearch-py-ml/blob/main/.github/workflows/integration.yml#L15) ?

dhrubo-os:
You can also run the whole test in your end like this: `./.ci/run-tests "opensearch" "true" 2.6.0` to check

AlibiZhenis:
It seems the test was failing because model unload was timing out. Once I increased the threshold to 60s, it's now passing on my side.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -63,6 +63,76 @@ def upload_model(
             model_path, model_config_path, isVerbose
         )
 
+    def upload_pretrained_model(
+        self, name: str, version: str, model_format: str, load_model: bool = True
+    ):
+        """
+        This method uploads a pretrained model into opensearch cluster using ml-common plugin's api.
+        First, this method creates a model id to store model info
+        and then loads the model from Hugging Face if load_model is True.
```
dhrubo-os, line 72: then loads the model in memory. We don't load model from huggingFace. Model is downloaded in their index and then it will be loaded into memory using DJL.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -63,6 +63,76 @@ def upload_model(
             model_path, model_config_path, isVerbose
         )
 
+    def upload_pretrained_model(
+        self, name: str, version: str, model_format: str, load_model: bool = True
```
dhrubo-os, line 67: how about `model_name`, `model_version`?

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -63,6 +63,76 @@ def upload_model(
...
+        This method uploads a pretrained model into opensearch cluster using ml-common plugin's api.
+        First, this method creates a model id to store model info
+        and then loads the model from Hugging Face if load_model is True.
+        The model has to be supported by ML Commons. Refer to https://opensearch.org/docs/latest/ml-commons-plugin/pretrained-models/.
+
+        :param name: Name of the pretrained model
+        :type name: string
+        :param version: Version of the pretrained model
+        :type version: string
+        :param model_format: Currently only supports "TORCH_SCRIPT"
```
dhrubo-os, line 79: We also support `ONNX`. If you see this : https://opensearch.org/docs/latest/ml-commons-plugin/pretrained-models/, we also have urls for `ONNX` artifacts.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -63,6 +63,76 @@ def upload_model(
...
+        :type model_format: string
+        :param load_model: Whether to load the model chunks from Hugging Face
```
dhrubo-os, line 81: whether to load the model in memory using uploaded model chunks.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -63,6 +63,76 @@ def upload_model(
...
+                raise Exception(output["error"])
+
+        return model_id
+
+    def _send_model_info(self, model_meta_json: dict):
+        """
+        This method sends the pretrained model info to ML Commons' upload api
+
+        :param model_meta_json: a dictionary object with model configurations
+        :type model_meta_json: string
```
dhrubo-os, line 113: dict instead of string

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -63,6 +63,76 @@ def upload_model(
...
+        :return: returns a unique id of the model
+        :rtype: int
```
dhrubo-os, line 115: string instead of int

dhrubo-os:
When you push a new commit with all these changes can you also run the workflow. Please let me know if you don't have permission to run workflow?

AlibiZhenis:
Sorry for all the mistakes. I'll try to be more attentive from now on.

It also seems that I don't have permission to run workflows.

tests/ml_commons/test_ml_commons_client.py
```diff
@@ -70,6 +74,47 @@ def test_init():
...
+        assert raised == False, "Raised Exception in getting pretrained model info"
+
+        raised = False
+        try:
+            ml_client.unload_model(model_id)
+            time.sleep(60)
+            ml_model_status = ml_client.get_model_info(model_id)
+            print("ml_model_status", ml_model_status)
+            assert ml_model_status.get("model_state") == "UNLOADED"
+        except:
```
dhrubo-os, line 106: Lint is failing for bare except, you need to add to skip linting error `# noqa: E722`

tests/ml_commons/test_ml_commons_client.py
```diff
@@ -70,6 +74,47 @@ def test_init():
...
+            assert ml_model_status.get("model_format") == "TORCH_SCRIPT"
+            assert ml_model_status.get("algorithm") == "TEXT_EMBEDDING"
+        except:  # noqa: E722
+            raised = True
+        assert raised == False, "Raised Exception in getting pretrained model info"
+
+        raised = False
+        try:
+            ml_client.unload_model(model_id)
+            time.sleep(60)
```
dhrubo-os, line 102: Integration is still failing. May be we can have a loop to check back in every 15 seconds. if we get the result, that's great otherwise will wait for another 15 seconds. 

But we can't run this forever. May be the hard limit can be 5 minutes. IF total duration passes more than 5 minutes then fail the test.

dhrubo-os:
> > So what should I do next? I have opensearch-py 2.2.0 installed, which pip says is the latest version.
> 
> Can you please change to `2.6.0` [here](https://github.com/opensearch-project/opensearch-py-ml/blob/main/.github/workflows/integration.yml#L15) ?

I saw you changed the version, but now i don't see the changed code. Where is it? Integration test is failing....

.github/workflows/integration.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         cluster: ["opensearch"]
         secured: ["true"]
         entry:
-          - { opensearch_version: 2.6.0 }
+          - { opensearch_version: 2.5.0 }
```
dhrubo-os: Why did you change this back?

AlibiZhenis: The tests are passing in my local environment, but not the workflow. So I wanted to see if the problem was with the version. Turns out it's not, I will change it back.

I am suspecting the problem is with the circuit breaker error. I am investigating.

AlibiZhenis:
It seems to be consistently passing now once I changed the version back to 2.6.0. I believe the reason is in my "minor code and test changes", where I changed the upload_pretrained_model method.

Now the method finishes once the model_state becomes loaded, instead of when the load model task gets completed.

dhrubo-os (approved):

b4sjoo (approved):

PR #119, opened 2023-03-22
Added Profile API

Merey1508:
### Description
These three methods all retrieve different types of profiles from an OpenSearch cluster using the ML Commons API. Here's a brief description of each method:

**def get_profile(self, node_ids: List[str] = [], return_all_tasks: bool = True, return_all_models: bool = True) -> object:**

This method retrieves a profile from an OpenSearch cluster for one or more specified node IDs.

**def get_profile_models(self, model_ids: List[str] = []) -> object:**

This method retrieves a profile from an OpenSearch cluster for one or more specified machine learning models.

**def get_profile_tasks(self, task_ids: List[str] = []) -> object:**

This method retrieves a profile from an OpenSearch cluster for one or more specified machine learning tasks.

All three methods are similar in structure and use the same API endpoint to retrieve different types of profiles based on the parameters passed to them.
 
### Issues Resolved
Link to issue: [#104 ](https://github.com/opensearch-project/opensearch-py-ml/issues/104)
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/opensearch_py_ml/ml_commons/ml_commons_client.py
+++ b/opensearch_py_ml/ml_commons/ml_commons_client.py
@@ -6,6 +6,7 @@
 # GitHub history for details.
 
 
+import json
 import time
 from typing import Any, List, Union
 
@@ -327,8 +328,8 @@ class MLCommonClient:
                 url=API_URL,
             )
 
-        if task_ids and len(task_ids) == 1:
-            API_URL = f"{ML_BASE_URI}/profile/tasks/{task_ids[0]}"
+        if task_ids and len(model_ids) == 1:
+            API_URL = f"{ML_BASE_URI}/profile/models/{model_ids[0]}"
             return self._client.transport.perform_request(
                 method="GET",
                 url=API_URL,
@@ -337,12 +338,21 @@ class MLCommonClient:
         API_URL = f"{ML_BASE_URI}/profile"
 
         API_BODY = {
-            "node_ids": node_ids,
-            "model_ids": model_ids,
-            "task_ids": task_ids,
             "return_all_tasks": return_all_tasks,
             "return_all_models": return_all_models,
         }
+
+        if len(node_ids) > 0:
+            API_BODY["node_ids"] = node_ids
+        if len(model_ids) > 0:
+            API_BODY["model_ids"] = model_ids
+        if len(task_ids) > 0:
+            API_BODY["task_ids"] = task_ids
+
+        try:
+            API_BODY = json.dumps(API_BODY)
+        except json.JSONDecodeError:
+            raise Exception("Invalid request body")
 
         return self._client.transport.perform_request(
             method="GET",
--- a/opensearch_py_ml/ml_commons/ml_commons_client.py
+++ b/opensearch_py_ml/ml_commons/ml_commons_client.py
@@ -294,3 +294,58 @@ class MLCommonClient:
             method="DELETE",
             url=API_URL,
         )
+
+    def get_profiles(
+        self,
+        node_ids: List[str] = [],
+        model_ids: List[str] = [],
+        task_ids: List[str] = [],
+        return_all_tasks: bool = True,
+        return_all_models: bool = True,
+    ) -> object:
+        """
+        This method retrieves the profile of one or more machine learning nodes, models, or tasks from OpenSearch cluster (using ml commons api)
+
+        :param node_ids: A list of node ids to retrieve profiles for (default: [])
+        :type node_ids: List[str]
+        :param model_ids: A list of model ids to retrieve profiles for (default: [])
+        :type model_ids: List[str]
+        :param task_ids: A list of task ids to retrieve profiles for (default: [])
+        :type task_ids: List[str]
+        :param return_all_tasks: a flag to indicate whether all tasks associated with the profiles should be returned (default: True)
+        :type return_all_tasks: bool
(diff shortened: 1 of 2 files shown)


Merey1508:
> * We need to add test for each methods to make sure the functionality.
> * We need to add documentation for each method in more details including describing parameters. Check other methods please.
> * Linting is failing. [check here](https://github.com/opensearch-project/opensearch-py-ml/blob/main/DEVELOPER_GUIDE.md#to-test-code-formatting-and-linting-issue-we-can-run)

Thanks for your comment. I fixed linting. Could you check the documentation? If I'm not mistaken, I added the documentation within code by shortly describing each method and their parameters.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -198,3 +198,76 @@ def delete_model(self, model_id: str) -> object:
             method="DELETE",
             url=API_URL,
         )
+
+    def get_profile(
```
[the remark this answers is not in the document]
Merey1508: Ok, I will try. Thank you for bringing this to my attention.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -198,3 +198,44 @@ def delete_model(self, model_id: str) -> object:
             method="DELETE",
             url=API_URL,
         )
+
+    def get_profiles(
```
dhrubo-os, line 202: we need to write corresponding test to make sure this is working or not.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -198,3 +198,44 @@ def delete_model(self, model_id: str) -> object:
...
+        :type task_ids: List[str]
+        :param return_all_tasks: a flag to indicate whether all tasks associated with the profiles should be returned (default: True)
+        :type return_all_tasks: bool
+        :param return_all_models: a flag to indicate whether all models associated with the profiles should be returned (default: True)
+        :type return_all_models: bool
+        :return: returns a json object containing the profile information of the specified nodes, models, or tasks from the OpenSearch cluster
+        :rtype: object
+        """
+
+        API_URL = f"{ML_BASE_URI}/profile"
```
dhrubo-os, line 227: Let's make this more robust so that we can also support scenarios like [here](https://opensearch.org/docs/latest/ml-commons-plugin/api/#profile)

1. GET /_plugins/_ml/profile/models --> If we see only one `model_id` is given nothing else, we can invoke this api. Example: `GET /_plugins/_ml/profile/models/model_id`
2. GET /_plugins/_ml/profile/tasks --> if we see only one `task_id` is given nothing else, we can invoke this api. 
Example: `GET /_plugins/_ml/profile/tasks/task_id`
3. Otherwise invoke the base API. Example:

```
GET /_plugins/_ml/profile
{
  "node_ids": ["KzONM8c8T4Od-NoUANQNGg"],
  "return_all_tasks": true,
  "return_all_models": true
}

```

Please let me know if you have any confusion. 

Also don't forget to write test. Let me know if you have any question regarding writing test.

dhrubo-os:
@Merey1508 is there any update on writing unit test? Please let me know if you need any help, thanks.

opensearch_py_ml/ml_commons/ml_commons_client.py
Nurlanprog: 1) Endpoint: GET /_plugins/_ml/profile/models/model_id  works fine.
2) Endpoint: GET /_plugins/_ml/profile/tasks/task_id  always returns empty json. So it was deleted. Instead for searching model tasks used API_BODY. 
3) In testing function for traversing over dictionary was added because model_id and task_id in returned json is 4 levels deeper.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -294,3 +295,67 @@ def delete_task(self, task_id: str) -> object:
...
+        """
+
+        if model_ids and len(model_ids) == 1:
+            API_URL = f"{ML_BASE_URI}/profile/models/{model_ids[0]}"
+            return self._client.transport.perform_request(
+                method="GET",
+                url=API_URL,
+            )
+
+        if task_ids and len(model_ids) == 1:
```
dhrubo-os, line 331: Can we try to add a comment to describe what exactly are we doing here?

opensearch_py_ml/ml_commons/ml_commons_client.py
rawwar: > 1. GET /_plugins/_ml/profile/models --> If we see only one `model_id` is given nothing else, we can invoke this api. Example: `GET /_plugins/_ml/profile/models/model_id`
> 2. GET /_plugins/_ml/profile/tasks --> if we see only one `task_id` is given nothing else, we can invoke this api.
>    Example: `GET /_plugins/_ml/profile/tasks/task_id`
> 3. Otherwise invoke the base API. Example:

@dhrubo-os : 
For 1, If we implement that, and then user wants all model profile data, its going to cause confusion. Example: User wants all tasks data and so, he doesn't pass model_id nor task_id . How are we going to conclude what user wanted here?

I think, a rather simple interface is to give users to call the respective api's separately.  we should have three separate methods each returning exactly what user is requesting.

dhrubo-os:
closing this pull request as this functionality is already done.

PR #126, opened 2023-03-28
Tests for mcorr libraries

greaa-aws:
### Description
This PR adds tests for the following tools used by metrics correlation:
- Haar discrete wavelet transformation:`wavelet_tools.py`
- Piecewise constant approximation and activity scoring: `wavelet_piecewise_const.py`
- Dip test for unimodality: `dip.py`
- Nonnegative matrix factorization with NNDSVD-A initialization: `nmf.py`
 
### Check List
  - [x] All tests pass

### DIFF
--- a/tests/ml_models/test_mcorr_dip_pytest.py
+++ b/tests/ml_models/test_mcorr_dip_pytest.py
@@ -22,6 +22,13 @@ import opensearch_py_ml.ml_models.metrics_correlation.dip as dip
 
 
 def test_dip_statistic():
+    '''
+    test correctness by computing the dip statistic on 
+    various input sequences. comparisons are made to the
+    equivalent output from the R package 'diptest', the 
+    canonical implementation of this method.
+    '''
+
     T = 128
 
     # uniform eCDF
@@ -54,6 +61,11 @@ def test_dip_statistic():
 
 
 def test_dip_pval():
+    '''
+    as above, reference values (this time for p-values)
+    are obtained from R package 'diptest'
+    '''
+
     # bimodal : p near zero
     T = 256
     xa = torch.arange(T + 1) / (T / 10)
@@ -148,8 +160,8 @@ def test_torchscript_output():
 
     tsD, tsP = loaded_model.forward(sig)
 
-    assert np.allclose(ptD, tsD)
-    assert np.allclose(ptP, tsP)
+    assert np.allclose(ptD, tsD, rtol=0, atol=0)
+    assert np.allclose(ptP, tsP, rtol=0, atol=0)
 
     os.remove("diptest.pt")
 
--- a/tests/ml_models/test_mcorr_dip_pytest.py
+++ b/tests/ml_models/test_mcorr_dip_pytest.py
@@ -1,0 +1,162 @@
+# SPDX-License-Identifier: Apache-2.0
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+# Any modifications Copyright OpenSearch Contributors. See
+# GitHub history for details.
+
+"""
+Tests of correctness and compilation for the dip test module
+of metrics correlation.
+
+"True" values here can be verified by comparison to established
+implementations such as the R package 'diptest'.
+"""
+
+import os
+
+import numpy as np
+import torch
+
+import opensearch_py_ml.ml_models.metrics_correlation.dip as dip
+
+
+def test_dip_statistic():
+    T = 128
+
+    # uniform eCDF
+    unifseq = torch.arange(T)
+    assert torch.abs(dip.dip(unifseq) - 0.003906) <= 1e-6
+
+    # uniform with large spike
+    unifspike = torch.arange(T)
+    unifspike[:50] = 20
+
+    assert torch.abs(dip.dip(unifspike) - 0.085426) <= 1e-6
+
+    # unimodal Gaussian
+    xs = torch.arange(-T // 2, T // 2 + 1)
+    x = torch.exp(-(xs**2) / 20)
+
+    assert torch.abs(dip.dip(x) - 0.010517) <= 1e-6
+
+    # max of 0.25 is (approx) achieved for (approx) bimodal data
+    apx_bimodal = torch.cat(
+        ((torch.normal(1.0, 0.001, [T // 2])), torch.normal(2.0, 0.001, [T // 2]))
+    )
+
+    assert dip.dip(apx_bimodal) >= 0.2475 and dip.dip(apx_bimodal) <= 0.25
+
+    # overlapping bimodal (mixture of Gaussians)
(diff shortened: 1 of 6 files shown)


tests/ml_models/test_mcorr_wavelet_pytest.py
```diff
@@ -0,0 +1,201 @@
...
+
+"""
+Tests of correctness and compilation for the wavelet module
+of metrics correlation.
+"""
+
+import os
+
+import numpy as np
+import pywt
```
dhrubo-os, line 16: we need to add this `pywt` in the `requirements-dev.txt` so that this plugin can be installed before testing. Otherwise integration test fails : https://github.com/opensearch-project/opensearch-py-ml/actions/runs/4548484040/jobs/8019564523?pr=126

```
 tests/ml_models/test_mcorr_wavelet_pytest.py:16: in <module>
    import pywt
E   ModuleNotFoundError: No module named 'pywt'
```

tests/ml_models/test_mcorr_dip_pytest.py
```diff
@@ -0,0 +1,162 @@
...
+
+import opensearch_py_ml.ml_models.metrics_correlation.dip as dip
+
+
+def test_dip_statistic():
+    T = 128
+
+    # uniform eCDF
+    unifseq = torch.arange(T)
+    assert torch.abs(dip.dip(unifseq) - 0.003906) <= 1e-6
```
dhrubo-os, line 29: Sometime test help us to understand the functionality of the code also.

From the `assert torch.abs(dip.dip(unifseq) - 0.003906) <= 1e-6`, I didn't understand why this assert is important and also the significance of these magic numbers?

May be we should add some comments to clarify the test intents here?

tests/ml_models/test_mcorr_dip_pytest.py
```diff
@@ -0,0 +1,162 @@
...
+
+    try:
+        _ = torch.jit.load("diptest.pt")
+    except Exception as exec:
+        assert False, f"Failed to load scripted model with exception {exec}"
+
+    os.remove("diptest.pt")
+
+
+def test_torchscript_output():
```
dhrubo-os, line 137: I think the idea of testing all these classes with conversion and without conversion.

tests/ml_models/test_mcorr_dip_pytest.py
greaa-aws: The objective of this test is to check the function that computes the dip statistic for correctness. As indicated in the comments at the top of the test file, this is done by comparing the output against a "gold standard" implementation on a set of relevant cases. 

The reason these magic numbers are hardcoded rather than generated on the fly is that the implementation is in R, not Python. It may be possible to call the R library from Python, but I'd need to do some investigation for how to make this work properly with the testing suite. Otherwise I can just add some more commentary here.

tests/ml_models/test_mcorr_nmf_pytest.py
```diff
@@ -0,0 +1,111 @@
...
+    ptW, ptH = model(X, 3, 200, 1e-6)
+
+    model = nmf.NMF()
+    jit_processing = torch.jit.script(model)
+    torch.jit.save(jit_processing, "nmftest.pt")
+    loaded_model = torch.jit.load("nmftest.pt")
+
+    tsW, tsH = loaded_model.forward(X, 3, 200, 1e-6)
+
+    assert torch.allclose(ptW, tsW)
```
dhrubo-os, line 108: although the comparison formula is similar, but i think in better to use `torch.testing.assert_close` which has more parameters open to check like `check_dtype`, `check_device`. [Ref](https://pytorch.org/docs/stable/testing.html#)

dhrubo-os:
When you are making the changes can you also please add the dict to list conversion of the output in this PR also. This will be helpful for me in mlcommons development. Thanks

tests/ml_models/test_mcorr_dip_pytest.py
dhrubo-os: > The objective of this test is to check the function that computes the dip statistic for correctness. As indicated in the comments at the top of the test file, this is done by comparing the output against a "gold standard" implementation on a set of relevant cases.
> 
> The reason these magic numbers are hardcoded rather than generated on the fly is that the implementation is in R, not Python. It may be possible to call the R library from Python, but I'd need to do some investigation for how to make this work properly with the testing suite. Otherwise I can just add some more commentary here.

Yeah, I thought so. I think for now to add more comment with be super helpful!!

tests/ml_models/test_mcorr_dip_pytest.py
```diff
@@ -0,0 +1,162 @@
...
+    ptD, ptP = model(sig)
+
+    model = DipTest()
+    jit_processing = torch.jit.script(model)
+    torch.jit.save(jit_processing, "diptest.pt")
+    loaded_model = torch.jit.load("diptest.pt")
+
+    tsD, tsP = loaded_model.forward(sig)
+
+    assert np.allclose(ptD, tsD)
```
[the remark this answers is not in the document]
greaa-aws: They aren't tensors: the function being tested here `diptest(seq: torch.Tensor) -> Tuple[float, float]`, so the outputs above are both floats. 

We can set the tolerances lower, but the defaults `(rtol = 1e-5, atol=1e-8)` already enforce a very tight "closeness" relative to the relevant scale of the outputs. I wonder if setting tolerances to exactly zero might cause tests to fail for numerical reasons not related to the correctness. This is why I have generally avoided exact vs `allclose()` type tests.

tests/ml_models/test_mcorr_nmf_pytest.py
greaa-aws: Good catch, and agreed. I'll change this everywhere.

tests/ml_models/test_mcorr_nmf_pytest.py
```diff
@@ -0,0 +1,111 @@
...
+            Wsk = sk_model.fit_transform(X.numpy())
+            Hsk = sk_model.components_
+
+            Wpt, Hpt = nmf_model(X, k=3, max_iter=iter, tol=1e-8)
+
+            W_rel_err = np.linalg.norm(Wpt.numpy() - Wsk) / np.linalg.norm(Wsk)
+            H_rel_err = np.linalg.norm(Hpt.numpy() - Hsk) / np.linalg.norm(Hsk)
+
+            assert W_rel_err <= 0.01
+            assert H_rel_err <= 0.01
```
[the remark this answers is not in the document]
greaa-aws: Sure, I'll add a comment here. The main reason is that we want to ensure that the NMF output matrices are highly similar overall to the gold standard method, which is a weaker condition than elementwise similarity. The numerical differences between pytorch and numpy can introduce small differences between elements that are not relevant for our purposes but can still make elementwise comparisons fail.

dhrubo-os:
linting is failing. Usually before pushing any changes I do : `nox -s lint`

dhrubo-os (approved):

PR #134, opened 2023-04-05
Model Conversion example notebooks

### LINKED ISSUE #123: [FEATURE] Convering traditional ML algorithms using Hummingbird and benchmark model performance.
Currently for build in ML algorithms in opensearch we need to write that in Java, which is sometimes more time consuming due to not having enough ML support in Java

One initiative we started is, we can write algorithm in [TorchScript](https://pytorch.org/docs/stable/jit.html), trace the torchScript file and then load the model file in Opensearch using MLCommon's Model serving framework.

One bottleneck is, [in torchScript we can't import any 3rd party library](https://pytorch.org/docs/stable/jit.html#python-functions-and-modules) like scikit-learn so to include scikit-learn models in Opensearch we have to rewrite the algorithm  in torchscrip which can't be the ideal solution.

To solve this problem we can use [Hummingbird](https://github.com/microsoft/hummingbird) through which we can convert traditional machine learning algorithm to neural network based algorithm for faster execution and at the same time we should be able to convert the algorithm to torchScript or Onnx so that we can load the model in opensearch.

In this issue, we would like to investigate if humming bird will solve our issue or not.

The following steps can be done in the investigation

1. Import Hummingbird in the py-ml repo
2. convert a model in torchScript (We can start with simple stateless models like [PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html)/[KernelPCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.KernelPCA.html#sklearn.decomposition.KernelPCA)
3. Run both format of the algorithm (original scikit learn algorithm and converted pytorch and onnx algorithm)to compare the output.
4. We can have bench-marking to compare performance between all three formats of model execution.

AlibiZhenis:
### Description
Currently, we can upload only TorchScript and ONNX models to OpenSearch. The problem is that we can't import any 3rd party libraries like scikit-learn. 

The aim of this PR was to investigate if we can convert sklearn models like Random Forest and PCA to TorchScript and ONNX in order to upload them to OpenSearch.

The jupyter notebooks in this PR serve as a tutorial on how to convert these models using Hummingbird and skl2onnx packages, and verify that converted models are actually identical to original ones through accuracy metrics.
 
### Issues Resolved
Closes #123 
 
### Check List


dhrubo-os (approved):

AlibiZhenis:
What about this PR?

dhrubo-os:
> 

Good call out. Can you please elaborate the description with more details about this PR (I know we can download the notebook to see, but still seeing detailed description can save reviewers time to review)

1. What is this experiment about
2. What did we do to do this experiment
3. What is the output of the experiment.

AlibiZhenis:
I added the description. Hope it clarifies all questions.

notebooks/model_conversion-pca.ipynb
```diff
@@ -0,0 +1,456 @@
...
+      "Do sklearn and torch KernelPCA models generate close results?\n",
+      "True\n",
+      "Do sklearn and torch TruncatedSVD models generate close results?\n",
+      "True\n"
+     ]
+    }
+   ],
+   "source": [
+    "for i, model_name in enumerate([\"PCA\", \"KernelPCA\", \"TruncatedSVD\"]):\n",
+    "    sklearn_transformed = sklearn_models[0].transform(X_test)\n",
```
greaa-aws, line 288: Should be `i`, not `0`?

notebooks/model_conversion-pca.ipynb
```diff
@@ -0,0 +1,456 @@
...
+    "    torch_transformed = torch_models[0].transform(X_test)\n",
```
greaa-aws, line 289: same as above

notebooks/model_conversion-pca.ipynb
```diff
@@ -0,0 +1,456 @@
...
+      "Do sklearn and onnx KernelPCA models generate close results?\n",
+      "True\n",
+      "Do sklearn and onnx TruncatedSVD models generate close results?\n",
+      "False\n"
+     ]
+    }
+   ],
+   "source": [
+    "for i, model_name in enumerate([\"PCA\", \"KernelPCA\", \"TruncatedSVD\"]):\n",
+    "    sklearn_transformed = sklearn_models[0].transform(X_test)\n",
```
greaa-aws, line 375: same as above

notebooks/model_conversion-pca.ipynb
```diff
@@ -0,0 +1,456 @@
...
+    "    \n",
+    "    sess = rt.InferenceSession(onnx_models[i].SerializeToString())\n",
+    "    input_name = sess.get_inputs()[0].name\n",
```
greaa-aws, line 378: same as above

notebooks/model_conversion-pca.ipynb
greaa-aws, line 375: Should be `i`, not `0`?

notebooks/model_conversion-classification.ipynb
```diff
@@ -0,0 +1,656 @@
...
+     "end_time": "2023-04-03T11:44:08.556536",
+     "exception": false,
+     "start_time": "2023-04-03T11:44:08.492198",
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "extra_config = {hb.operator_converters.constants.BATCH_SIZE: 1}\n",
```
greaa-aws, line 341: could you explain why this is necessary for the knn classifier?

notebooks/model_conversion-classification.ipynb
```diff
@@ -0,0 +1,656 @@
...
+     "duration": 0.008582,
+     "end_time": "2023-04-03T11:44:07.980088",
+     "exception": false,
+     "start_time": "2023-04-03T11:44:07.971506",
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "source": [
+    "# Convert the models to torchscript using Hummingbird and test the results (visualized further below)"
```
greaa-aws, line 244: why are these results visualized if they're just scalar values? can't we just check that the metrics are equal, as in the notebook below?

notebooks/model_conversion-classification.ipynb
AlibiZhenis: Otherwise, the following exception is raised:

`RuntimeError: Hummingbird requires explicit specification of batch_size parameter when compiling KNeighborsClassifier`

notebooks/model_conversion-classification.ipynb
AlibiZhenis: You're right. But since we have 3 algorithms, 3 different models for each of them, and 3 metrics for each one too (accuracy, precision, recall), visualizing the results seemed like the more convenient choice. 

In PCA notebooks, since we don't have metrics to measure PCA results, I just compared pairwise elements of the matrices.

In this case, it would require too many calls to isclose(), 27 to be exact.

AlibiZhenis: We can reduce it to 9 if we make an array of metrics for each algorithm.

dhrubo-os:
@AlibiZhenis Integration test is failing. Can you please check?

AlibiZhenis:
test_save_as_onnx() failed. I haven't touched it. Maybe it's because of added dependencies, though I doubt it. Can you run the tests again?

dhrubo-os:
> test_save_as_onnx() failed. I haven't touched it. Maybe it's because of added dependencies, though I doubt it. Can you run the tests again?

Sure, I'll do that. Can you also please address the other comments?

dhrubo-os:
> 



> test_save_as_onnx() failed. I haven't touched it. Maybe it's because of added dependencies, though I doubt it. Can you run the tests again?

Looks like all checks passed successfully. Can you address/reply to the remaining comments? Thanks.

AlibiZhenis:
I addressed the other comments 3 weeks ago. Can you also reply to the other PR?

notebooks/model_conversion-pca.ipynb
dhrubo-os: @AlibiZhenis I was referring these comments.

AlibiZhenis: I think these comments were already addressed [here](https://github.com/opensearch-project/opensearch-py-ml/pull/134/commits/f2b1cb09ef470e9caf7c1940edf6dc2f089f291c)

greaa-aws:
High-level comments:

1. Why do we have two different notebooks (`model_conversion-pca.ipynb` and `model_conversion-classification.ipynb`) to demonstrate the same model conversion functionality? If we want to show model conversion for multiple kinds of `sklearn` model, this can easily be done in a single notebook. 

2. Right now, the only test we apply after converting `sklearn` models to other formats is that the pre- and post-conversion models have similar predictions. Can we also show a test that confirms the post-conversion model has the appropriate format, and can be manipulated as such? For example, check that we can save and load the Hummingbird-converted TorchScript models using the `torch.jit` module?

3. Demo notebooks should have more of a narrative structure, so that users can easily follow along rather than just read a bunch of code. At a minimum I think this involves:
    - An “Overview” or “Introduction” section that describes the purpose of the notebook, walks through the main steps, and clearly states what a reader should learn / understand by the end.
    - A text block before each code block that contains at least a short statement of what we’re doing next.
    - A conclusion that summarizes the main result and points users to the full documentation if they want to learn more.

notebooks/model_conversion-classification.ipynb
greaa-aws: I disagree. Plots are not the right way to compare 3 scalar values, even if that comparison happens multiple times. The overall number of calls to `allclose()` here is really not that much. You could write a function that runs comparisons given a list of models and metrics, returning `True` if all are close and `False` otherwise, and apply that. Alternatively, you could summarize the results in a `pd.DataFrame`.

notebooks/model_conversion-pca.ipynb
```diff
@@ -0,0 +1,427 @@
...
+     "exception": false,
+     "start_time": "2023-04-03T08:30:21.087974",
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "for model in sklearn_models:\n",
+    "    model.fit(X_train)"
```
greaa-aws, line 165: Right now this is throwing a not-converged warning for the SVM classifier. While training convergence doesn't actually matter for the purposes of this notebook, it might be good to add a brief text comment here acknowledging that.

dblock:
Closing as unfinished, please feel free to finish and reopen per comments above. Thanks!

[Catch All Triage w/ [1](https://github.com/dblock), [2](https://github.com/krisfreedain), [3](https://github.com/andrross)]

PR #144, opened 2023-04-15
Added search task API

Nurlanprog:
### Description
User can search task by providing one by one arguments to function or by providing manually json with parameters of task. Test is added.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-py-ml/issues/104
 
### Check List
  - [ ] All tests pass


opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -194,6 +195,36 @@ def _get_task_info(self, task_id: str):
...
+        """
+
+        API_URL = f"{ML_BASE_URI}/tasks/_search"
+
+        query_bool = {"query": {"bool": {"filter": []}}}
+
+        API_BODY = query_bool
+
+        for key, value in kwargs.items():
+            if key == "json":
```
dhrubo-os, line 214: are we expecting somebody to mention "json" in their provided value? Can you give example for this what are you expecting customer to provide?

tests/ml_commons/test_search_task.py
```diff
@@ -0,0 +1,34 @@
...
+# Any modifications Copyright OpenSearch Contributors. See
+# GitHub history for details.
+
+from opensearch_py_ml.ml_commons import MLCommonClient
+from tests import OPENSEARCH_TEST_CLIENT
+
+ml_client = MLCommonClient(OPENSEARCH_TEST_CLIENT)
+
+
+def test_search_task():
```
dhrubo-os, line 14: I think in this case test needs to be more detailed. Right now we are just testing if we get 200 status or not. But what if for a bad query we didn't find any result what we supposed to but still we get 200 status, that will not satisfy the requirement of the test. 

So, I would suggest to do test from the perspective of what we are expecting to get with this search result and what is in reality. To do that even if we need to ingest some model or task, we need to do that to get the search result.

opensearch_py_ml/ml_commons/ml_commons_client.py
Nurlanprog: Yes, When calling function if user provides named parameter like this: search_task(json={...})

tests/ml_commons/test_search_task.py
Nurlanprog: Ok, I got it. I will try to upload, load model. Then I will try to search based on new task.

dhrubo-os: we are already uploading model and loading models, can we try to use that?

opensearch_py_ml/ml_commons/ml_commons_client.py
dhrubo-os: How about we make this more intuitive rather than asking customer to provide `json`. Can we do something like this:

```
def my_function(**kwargs):
    for key, value in kwargs.items():
        if isinstance(value, str):
            try:
                value = json.loads(value.replace("'", '"'))
                kwargs[key] = value
            except json.JSONDecodeError:
                pass
        print(f"{key}: {value}")
```

You need to change the code based on your requirement. But what I meant is, if we see somebody is providing a string, then we can try to load that in a json format.

Nurlanprog: Ok, I understood. I didn't take into account the convenience. I will change the code.

tests/ml_commons/test_search_task.py
Nurlanprog: As I understood, you mean I should use upload, load stuff from test_ml_commons file.

dhrubo-os: Yeah, for example, when a model is uploaded in the`test_ml_commons` file, a task will be created. We get the task id.

Now we can also search using this search task api to get the same task id, then we can assert that both task ids are same. That can be a good testing that search task api is really providing results.

What do you think?

Nurlanprog: Yeah, good idea. I thought about it, but I wrote the test in another file because I thought that search task is not intended to be in function with upload, load stuff (because purpose is different).

opensearch_py_ml/ml_commons/ml_commons_client.py
Nurlanprog: I believe that I already did it. It's in else block. Providing json is just one option of setting API_BODY. User can also provide just key and according values. Finally, I used json.dumps for formatting, if in example user provided single quote instead of double quote.

tests/ml_commons/test_search_task.py
Nurlanprog: Just to warn, I have raised PR for integrating search_model also. It has pretty similar code. I will change test code for search model also.

opensearch_py_ml/ml_commons/ml_commons_client.py
dhrubo-os: and also we should throw `except json.JSONDecodeError:` if the json format is not right. before sending the request to api endpoint. 

When you make the changes, please also describe in the PR description what are the ways (what did you send in `kwargs`) you tried to search using your function.

Nurlanprog: Ok, I will add code for checking json.

In the description of PR, I mentioned all ways of providing parameters to function. Sorry if it is not so much descriptive.

dhrubo-os: I meant in your testing for the functionality what parameters you provided to test.  List down all of them, then I can suggest may be other ways to test the functionality.

Nurlanprog: Ok, I got it. Thank you.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -194,6 +195,36 @@ def _get_task_info(self, task_id: str):
...
+        :rtype: object
+        """
+
+        API_URL = f"{ML_BASE_URI}/tasks/_search"
+
+        API_BODY = {"query": {"bool": {"filter": []}}}
+
+        for key, value in kwargs.items():
+            try:
+                if key == "json":
```
dhrubo-os, line 213: We don't expect user's to provide any key with `json`

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -194,6 +195,36 @@ def _get_task_info(self, task_id: str):
...
+                    API_BODY = value
+                else:
+                    term_query = {"term": {key: value}}
+                    API_BODY["query"]["bool"]["filter"].append(term_query)
```
dhrubo-os, line 217: What happens if user want's to run any `should` query?

tests/ml_commons/test_ml_commons_client.py
```diff
@@ -183,6 +183,13 @@ def test_integration_model_train_upload_full_cycle():
                     print("Model Task Status:", ml_task_status)
                     raised = True
                 assert raised == False, "Raised Exception in pulling task info"
+
+                try:
+                    search_task_obj = ml_client.search_task(_id=task_id)
```
dhrubo-os, line 188: can we try to search in other ways also?

In testing, we usually try to see how other ways we can search to make sure this search_task is functioning accordingly.

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -359,6 +360,36 @@ def _get_task_info(self, task_id: str):
             url=API_URL,
         )
 
+    def search_task(self, **kwargs) -> object:
```
dhrubo-os, line 363: Sorry for making this complicated for you. Let's make this completely simple. 

1. We will expect a json string, similar like [this](https://opensearch.org/docs/2.4/ml-commons-plugin/api/#search-task) which will be a query body.

2. we will check if the query body is a valid json or not. We can check this :

import json

```
def is_json(myjson):
  try:
    json.loads(myjson)
  except ValueError as e:
    return False
  return True
```
3. And if the json is valid then we will simple call the api: `GET /_plugins/_ml/tasks/_search`.

Let's keep this simple.

I wanted to make this completely dynamic so that customer doesn't need to build the query by himself, but it seems like it can be bit complex and we might not cover all the edge cases. So better to expect a valid search query body to make things simple.

Nurlanprog: Hello. Sorry for dragging out this task for so long. I was worked in my bachelor diploma work. Now, I am free and I am already working on it. As promised I will do this task.

dhrubo-os: No worries. Thanks for working on it. Please let me know if you have any question or concern.

Nurlanprog: I will try to finish before Sunday and make it dynamic so customer does not need to build the query by himself. Anyway if it will not work I will just make just json input for customer from scratch.

dhrubo-os: Sounds good. May be you can take a look at [this article](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-your-data.html#common-search-options) to understand more options about the search in elastic search or opensearch.

Nurlanprog: I collected all url parameters and main request body parameters of search api. I made helper functions for building term query, range, match, bool query, geo, term level queries. Can I finish collecting all search api request body parameters till thursday with testing?

Nurlanprog: Example of helper function that I made: 
def range_query(field, gte=None, gt=None, lte=None, lt=None):
    query = {"range": {field: {}}}
    if gte is not None:
        query["range"][field]["gte"] = gte
    if gt is not None:
        query["range"][field]["gt"] = gt
    if lte is not None:
        query["range"][field]["lte"] = lte
    if lt is not None:
        query["range"][field]["lt"] = lt
    return query

dhrubo-os: I feel like to cover all the use cases in a single method, we might need to have lot of parameters and this might get confusing for users also. May be we should look for the simpler solution as this query pattern is pretty well used by users? What do you think?

Nurlanprog: Ok

Nurlanprog: Function and testing is done. From syncing with latest changes I have several "merge branch" commits. Is it ok?

Nurlanprog: I will rise PR again. I broke rebasing.

PR #158, opened 2023-05-03
Allow ml-models/model_listing to publish without version

prudhvigodithi:
### Description
Allow ml-models/model_listing model to bypass the version
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/jenkins/ml-models.JenkinsFile
+++ b/jenkins/ml-models.JenkinsFile
@@ -28,7 +28,7 @@ pipeline {
         )
         }
     environment {
-        ARTIFACT_PATH = "${BASE_DOWNLOAD_PATH}/${VERSION}/"
+        ARTIFACT_PATH = "${env.BASE_DOWNLOAD_PATH == "ml-models/model_listing/pre_trained_models.json" ? "${env.BASE_DOWNLOAD_PATH}" : "${env.BASE_DOWNLOAD_PATH}/${env.VERSION}/"}"
         UPLOAD_PATH = "models/ml-models"
         BUCKET_NAME = credentials('ml-models-bucket-name')
     }
@@ -36,11 +36,19 @@ pipeline {
         stage('Parameters Check') {
             steps {
                 script {
-                    if(BASE_DOWNLOAD_PATH.isEmpty() || VERSION.isEmpty()) {
+                    if(env.BASE_DOWNLOAD_PATH == "ml-models/model_listing/pre_trained_models.json" && env.VERSION.isEmpty()) {
+                        echo "Proceeding to download the ml-models/model_listing."
+                    }                  
+                    else if (env.BASE_DOWNLOAD_PATH.isEmpty()) {
                         currentBuild.result = 'ABORTED'
                         error('Parameters cannot be empty! Please provide the correct values.')
+                    } else if (env.VERSION.isEmpty()) {
+                        currentBuild.result = 'ABORTED'
+                        error('Parameters cannot be empty! Please provide the correct values.')
+                    } else {
+                        echo "Proceeding to download the ml-models."
                     }
-                    if(BASE_DOWNLOAD_PATH.endsWith('/')) {
+                    if(env.BASE_DOWNLOAD_PATH.endsWith('/')) {
                         currentBuild.result = 'ABORTED'
                         error('"/" not allowed at the end of the BASE_DOWNLOAD_PATH') 
                     }
@@ -53,9 +61,9 @@ pipeline {
                     downloadFromS3(
                         assumedRoleName: 'get_models',
                         roleAccountNumberCred: 'ml-models-aws-account-number',
-                        downloadPath: "${ARTIFACT_PATH}",
-                        bucketName: "${BUCKET_NAME}",
-                        localPath: "${WORKSPACE}/artifacts",
+                        downloadPath: "${env.ARTIFACT_PATH}",
+                        bucketName: "${env.BUCKET_NAME}",
+                        localPath: "${env.WORKSPACE}/artifacts",
                         force: true,
                         region: 'us-west-2'
                     )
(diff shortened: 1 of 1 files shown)


jenkins/ml-models.JenkinsFile
```diff
@@ -28,17 +28,25 @@ pipeline {
         )
         }
     environment {
-        ARTIFACT_PATH = "${BASE_DOWNLOAD_PATH}/${VERSION}/"
+        ARTIFACT_PATH = "${env.BASE_DOWNLOAD_PATH == "ml-models/model_listing" ? "${BASE_DOWNLOAD_PATH}" : "${BASE_DOWNLOAD_PATH}/${VERSION}/"}"
```
ylwu-amzn, line 31: Should we add `.json` suffix?

gaiksaya, line 31: Did not get why `env.` was added

prudhvigodithi: All the files part of `ml-models/model_listing` would be published @gaiksaya @peterzhuamazon correct me if i'm wrong.
Thank you

jenkins/ml-models.JenkinsFile
```diff
@@ -28,17 +28,25 @@ pipeline {
...
         BUCKET_NAME = credentials('ml-models-bucket-name')
     }
     stages{
         stage('Parameters Check') {
             steps {
                 script {
-                    if(BASE_DOWNLOAD_PATH.isEmpty() || VERSION.isEmpty()) {
+                    if(BASE_DOWNLOAD_PATH == "ml-models/model_listing" && VERSION.isEmpty()) {
+                        echo "Proceeding to download the ml-models/model_listing."
+                    }
```
gaiksaya, line 41: Lots of if and else can we have a better conditions?
Also you can ignore last else block. No need to log normal flow.

```suggestion
                    if(BASE_DOWNLOAD_PATH == "ml-models/model_listing" && !VERSION.isEmpty()){
                        currentBuild.result = 'ABORTED'
                        error('ml-models/model_listing cannot have a version')
                    }                  
```

jenkins/ml-models.JenkinsFile
gaiksaya: I think we should add the suffix if it is just 1 file to avoid downloading and uploading any other files as a security measure. WDYT?

prudhvigodithi: That is actual way in Jenkins to use the environment variables.

jenkins/ml-models.JenkinsFile
prudhvigodithi: We need all those cases Sayali, removing a condition can cause a run to have a mismatch of parameters.
1) Check for base path as `ml-models/model_listing` and if so version should be empty.
2) Check for empty base path.
3) Check for empty version.
4) This is an else block, to allow move to next stage, hence cant be removed as its an else in if-else condition, not just an if condition.

With the above 4 conditions I have tested the following.
1) With base path as `ml-models/model_listing`, user cant have an version, if version is declared raise an error.
2) User can run the pipeline with an empty base path nor with an empty version value, unless the base path is `ml-models/model_listing`

jenkins/ml-models.JenkinsFile
gaiksaya: Yeah my question was where are we adding `BASE_DOWNLOAD_PATH` as env variable. Are parameters by default env variables?

jenkins/ml-models.JenkinsFile
gaiksaya: Thanks! Wanted to confirm the 1st point which works I believe. 
My point was the checks are only to abort the builds and do nothing if everything is ok

jenkins/ml-models.JenkinsFile
prudhvigodithi: Ok here is some information.
`env` contains all environment variables.
`params` contains all build parameters.
Jenkins also automatically creates an environment variable for each build parameter. Hence be accessed with `env.`

jenkins/ml-models.JenkinsFile
prudhvigodithi: `do nothing if everything is ok`, in an else block i'm forced to move forward and do nothing, hence added an `echo` statement, we cant skip an else block section.

prudhvigodithi: Just to add with your suggestion commit, I believe the job passes if user just declared a `VERSION` parameter without `BASE_DOWNLOAD_PATH`, which I have handled in if-else.

jenkins/ml-models.JenkinsFile
prudhvigodithi: Sorry missed this, yes good point.

gaiksaya: Thanks that answers it Can you make it consistent across the jenkins file then instead of just one instance? Would help to avoid confusion

prudhvigodithi: Just one point I believe there can be multiple JSON files under `ml-models/model_listing` that should be pushed without version, in that case adding suffix for 1 file would be the right solution and entire path should be allowed `ml-models/model_listing` @dhrubo-os will we have multiple files under `ml-models/model_listing` in future?

dhrubo-os: May be for now we can point to the 1 file we are talking about here so that we don't release any other files accidentally. I'm not 100% sure if we will add another json file here or not. 

So may be we can update this code later if that requirement appears later. What do you think?

prudhvigodithi: Thanks pointed to one file`ml-models/model_listing/pre_trained_models.json`.

prudhvigodithi: Sure thanks

gaiksaya (approved):

dhrubo-os (approved):

PR #191, opened 2023-07-14
Make `make_model_config_json` function more concise

thanawan-atc:
### Description
Instead of scrapping `embedding_dimension`, `pooling_mode`, and `normalize_result` from config.json file, we can use  sentence transformer attributes and build-in function directly.

### Check List

### DIFF
--- a/opensearch_py_ml/ml_models/sentencetransformermodel.py
+++ b/opensearch_py_ml/ml_models/sentencetransformermodel.py
@@ -1080,7 +1080,7 @@ class SentenceTransformerModel:
                 normalize_result = True
             else:
                 normalize_result = False
-                
+
         model_config_content = {
             "name": model_name,
             "version": version_number,
@@ -1091,7 +1091,7 @@ class SentenceTransformerModel:
                 "embedding_dimension": embedding_dimension,
                 "framework_type": "sentence_transformers",
                 "pooling_mode": pooling_mode,
-                "normalize_result": normalize_result
+                "normalize_result": normalize_result,
                 "all_config": json.dumps(all_config),
             },
         }
--- a/opensearch_py_ml/ml_models/sentencetransformermodel.py
+++ b/opensearch_py_ml/ml_models/sentencetransformermodel.py
@@ -24,6 +24,7 @@ import torch
 import yaml
 from accelerate import Accelerator, notebook_launcher
 from sentence_transformers import SentenceTransformer
+from sentence_transformers.models import Pooling, Normalize, Transformer
 from torch.utils.data import DataLoader
 from tqdm import tqdm
 from transformers import TrainingArguments, get_linear_schedule_with_warmup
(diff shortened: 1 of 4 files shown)


opensearch_py_ml/ml_models/sentencetransformermodel.py
```diff
@@ -1093,61 +1091,15 @@ def make_model_config_json(
...
-                        ". Please check the config.json ",
-                        "file in the path.",
-                    )
-
         if normalize_result is not None:
             model_config_content["model_config"]["normalize_result"] = normalize_result
         else:
-            normalize_result_json_file_path = os.path.join(folder_path, "2_Normalize")
-            if os.path.exists(normalize_result_json_file_path):
+            if len(model._modules) >= 3 and isinstance(model._modules["2"], Normalize):
```
dhrubo-os, line 1102: I think otherwise we can say `normalize_result` = False?

opensearch_py_ml/ml_models/sentencetransformermodel.py
```diff
@@ -1028,8 +1028,37 @@ def make_model_config_json(
...
+        model = SentenceTransformer(self.model_id)
+        if model_type is None:
+            if len(model._modules) >= 1 and isinstance(
+                model._modules["0"], Transformer
+            ):
+                try:
+                    model_type = model._modules["0"].auto_model.__class__.__name__
+                    model_type = model_type.lower().rstrip("model")
+                except Exception as e:
+                    raise Exception(f"Raised exception while getting model_type: {e}")
```
dhrubo-os, line 1042: Let's try to cover these exception branches. I think either we can mock the model or we can stub the `_modules` as None so that it can go to the except branch. Same for other exceptions.

thanawan-atc: I don't think we can do it that way because the function loads a hugging face SentenceTransformer within the function, so we cannot stub the _modules as None in test file.

opensearch_py_ml/ml_models/sentencetransformermodel.py
thanawan-atc: Edited!

opensearch_py_ml/ml_models/sentencetransformermodel.py
```diff
@@ -1028,8 +1022,43 @@ def make_model_config_json(
         if model_name is None:
             model_name = self.model_id
 
-        # if user input model_type and embedding_dimension, it will skip reading the config.json file
-        if model_type is None or embedding_dimension is None:
+        # if user input model_type/embedding_dimension/pooling_mode, it will skip this step.
+        model = SentenceTransformer(self.model_id)
```
dhrubo-os, line 1026: can we use `model_name` here? Customer might provide another model name here

opensearch_py_ml/ml_models/sentencetransformermodel.py
dhrubo-os: I agree this will be bit tricky to implement. We can skip this for now. Not urgent.

opensearch_py_ml/ml_models/sentencetransformermodel.py
thanawan-atc: I can change it to be `model_name`, but `model_name` still needs to be hugging face `model_id` for it to work though (i.e., 'sentence-transformers/msmarco-distilbert-base-tas-b'). 

I was also wondering why we need to have model_name. I thought that we just want to allow user to set the arbitrary name for the model. But in that case line 1026 should still be `self.model_id` because `SentenceTransformer("Arbitrary Name")` would not work [Note that it is hugging face class `SentenceTransformer` not our class ` SentenceTransformerModel` here.] I'm not sure if I misunderstand anything. Let me know if you still want me to change it to `model_name`.

dhrubo-os: You were right. We can't use `model_name`. For some reason I mixed up with `model_name` with `model_id`. Ignore my comment.

dhrubo-os (approved):

rbhavna (approved):

PR #198, opened 2023-07-21
[Draft for Initial Review] Automating Model Autotracing and Uploading Process

thanawan-atc:
### Description
[Describe what this change achieves]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


.github/workflows/model_uploader.yml
```diff
@@ -0,0 +1,321 @@
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-region: ${{ secrets.MODEL_UPLOADER_AWS_REGION }}
+          role-to-assume: ${{ secrets.MODEL_UPLOADER_ROLE }}
+          role-session-name: model-auto-tracing
+      - name: Dryrun model uploading
+        id: dryrun_model_uploading
+        run: |
+          aws s3 sync ./upload/ s3://opensearch-exp/ml-models/huggingface/sentence-transformers/ --dryrun
```
dhrubo-os, line 123: 1. `opensearch-exp` don't we need to replace this with `opensearch-models`
2. I was thinking `huggingface/sentence-transformers` will be set from the model id we are providing to keep the flow more generic? Or at least for now let's keep `sentence-transformers` dynamic. For example: if we give model id as `sentence-transformers/msmarco-distilbert-base-tas-b` then it will create a file path for : 
`opensearch-models/ml-models/huggingface/sentence-transformers/msmarco-distilbert-base-tas-b`. And for `meta-llama/Llama-2-7b-chat-hf` it'll create file path: `opensearch-models/ml-models/huggingface/meta-llama/Llama-2-7b-chat-hf`.

.github/workflows/model_uploader.yml
```diff
@@ -0,0 +1,321 @@
...
+          dryrun_output=$(aws s3 sync ./upload/ s3://opensearch-exp/ml-models/huggingface/sentence-transformers/ --dryrun)
+          echo "dryrun_output<<EOF" >> $GITHUB_OUTPUT
+          echo "${dryrun_output@E}" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+          echo "${dryrun_output@E}"
+    outputs:
+      dryrun_output: ${{ steps.dryrun_model_uploading.outputs.dryrun_output }}
+
+  # Step 4: Ask for manual approval from the CODEOWNERS
+  manual-approval:
```
dhrubo-os, line 133: What happens if we realized there's a mistake and we don't want to approve?

.github/workflows/model_uploader.yml
```diff
@@ -0,0 +1,321 @@
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-region: ${{ secrets.MODEL_UPLOADER_AWS_REGION }}
+          role-to-assume: ${{ secrets.MODEL_UPLOADER_ROLE }}
+          role-session-name: model-uploading
+      - name: Copy Files to the Bucket
+        id: copying_to_bucket
+        run: |
+          aws s3 sync ./upload/ s3://opensearch-exp/ml-models/huggingface/sentence-transformers/
```
dhrubo-os, line 204: if we can keep this value in a variable that way we could reuse the same value.

.github/workflows/model_uploader.yml
```diff
@@ -0,0 +1,321 @@
...
+          
+          ========= Workflow Details ==========
+          - Workflow Name: ${{ github.workflow }}
+          - Workflow Run ID: ${{ github.run_id }}
+          - Workflow Initiator: @${{ github.actor }}
+          
+          ========= Model Information =========
+          - Model ID: ${{ github.event.inputs.model_id }}
+          - Model Version: ${{ github.event.inputs.model_version }}
+          - Tracing Format: ${{ github.event.inputs.tracing_format }}
```
dhrubo-os, line 257: How will this support "BOTH" case? As we discussed we need to have two separate objects if the tracing option is "Both"

noxfile.py
```diff
@@ -142,10 +142,26 @@ def test(session, pandas_version: str):
...
-
     session.install("-r", "docs/requirements-docs.txt")
     session.install(".")
 
     session.cd("docs")
     session.run("make", "clean", external=True)
     session.run("make", "html", external=True)
+
+
+@nox.session
```
dhrubo-os, line 153: Let's add comment here for developers not to get confused.

utils/model_uploader/model_autotracing.py
```diff
@@ -0,0 +1,452 @@
...
+        assert (
+            False
+        ), f"Raised Exception in tracing {model_format} model\
+                             during initiating a sentence transformer model class object: {e}"
+
+    # 2.) Save the model in the specified format
+    model_path = None
+    try:
+        if model_format == TORCH_SCRIPT_FORMAT:
+            model_path = pre_trained_model.save_as_pt(
```
dhrubo-os, line 95: How are we planning to add the LICENSE file?

utils/model_uploader/model_autotracing.py
```diff
@@ -0,0 +1,452 @@
...
+    model_id = ""
+    try:
+        model_id = ml_client.register_model(
+            model_path=model_path,
+            model_config_path=model_config_path,
+            deploy_model=True,
+            isVerbose=True,
+        )
+        print()
+        print(f"{model_format}_model_id:", model_id)
```
dhrubo-os, line 153: Can't we add \n here and then remove the `print()`?

utils/model_uploader/model_autotracing.py
```diff
@@ -0,0 +1,452 @@
...
+    except Exception as e:
+        assert (
+            False
+        ), f"Raised Exception in {model_format} model registration/deployment: {e}"
+
+    # 2.) Check model status
+    try:
+        ml_model_status = ml_client.get_model_info(model_id)
+        print()
+        print("Model Status:")
```
dhrubo-os, line 164: `print("\n Model Status: ", ml_model_status)`?

utils/model_uploader/model_autotracing.py
```diff
@@ -0,0 +1,452 @@
...
+        print(ml_model_status)
+        assert ml_model_status.get("model_format") == model_format
+        assert ml_model_status.get("algorithm") == "TEXT_EMBEDDING"
```
dhrubo-os, line 167: We need to check if model is DEPLOYED or not

utils/model_uploader/model_autotracing.py
```diff
@@ -0,0 +1,452 @@
...
+    except Exception as e:
+        assert (
+            False
+        ), f"Raised Exception in generating sentence embedding with {model_format} model: {e}"
+
+    # 4.) Undeploy the model
+    try:
+        ml_client.undeploy_model(model_id)
+        ml_model_status = ml_client.get_model_info(model_id)
+        assert ml_model_status.get("model_state") != "UNDEPLOY_FAILED"
```
dhrubo-os, line 188: What other status are we getting here? Can't we test if the model status is "undeployed" or not?

utils/model_uploader/model_autotracing.py
```diff
@@ -0,0 +1,452 @@
...
+                rtol=RTOL_TEST,
+                atol=ATOL_TEST,
+            )
+        except Exception as e:
+            failed_cases.append((TEST_SENTENCES[i], e))
+
+    if len(failed_cases):
+        print()
+        print(
+            f"Original embeddings DOES NOT matches {model_format} embeddings in the following case(s):"
```
dhrubo-os, line 233: I think i would like to add the model id information in the output as well. In that case, rather than sending `model_format` and `model_id` in this function, we can just send true/false from this function as return value and then handle the exception from the parent main function.

.github/workflows/model_uploader.yml
thanawan-atc: Will make those changes!

.github/workflows/model_uploader.yml
```diff
@@ -0,0 +1,321 @@
...
+      - name: Dryrun model uploading
+        id: dryrun_model_uploading
+        run: |
+          aws s3 sync ./upload/ s3://opensearch-exp/ml-models/huggingface/sentence-transformers/ --dryrun
+          dryrun_output=$(aws s3 sync ./upload/ s3://opensearch-exp/ml-models/huggingface/sentence-transformers/ --dryrun)
+          echo "dryrun_output<<EOF" >> $GITHUB_OUTPUT
+          echo "${dryrun_output@E}" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+          echo "${dryrun_output@E}"
+    outputs:
```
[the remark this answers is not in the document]
thanawan-atc: Yes. It will show where each file gets uploaded to.

.github/workflows/model_uploader.yml
thanawan-atc: We can comment `deny` or `no` in the issue.

.github/workflows/model_uploader.yml
thanawan-atc: Will look into that. From what I know right now is that the variable in each `job` (i.e., model autotracing / manual approval / ...) is independent.

.github/workflows/model_uploader.yml
thanawan-atc: It will raise one PR, but there will be two objects in `support_models.json` and `MODEL_UPLOAD_HISTORY.md`.

noxfile.py
thanawan-atc: Ok!

utils/model_uploader/model_autotracing.py
thanawan-atc: It's added in `prepare_files_for_uploading` step.

utils/model_uploader/model_autotracing.py
thanawan-atc: Sure

utils/model_uploader/model_autotracing.py
thanawan-atc: Screenshot 2023-07-24 at 11 32 52 AM

Doesn't this line already handle that?

utils/model_uploader/model_autotracing.py
thanawan-atc: Let me look into that. I follow the integration test here.

utils/model_uploader/model_autotracing.py
thanawan-atc: Sure!

.github/workflows/model_uploader.yml
dhrubo-os: What'll happen if we comment `deny` or `no` in the issue?

utils/model_uploader/model_autotracing.py
dhrubo-os: Technically I agree. But in the `Check model status` status asserting model status kinda makes more sense to me. Also what'll happen to the workflow if any of this code throws the exception?

utils/model_uploader/update_models_upload_history_md.py
```diff
@@ -0,0 +1,246 @@
...
+    models = []
+    if os.path.exists(MODEL_JSON_FILEPATH):
+        with open(MODEL_JSON_FILEPATH, "r") as f:
+            models = json.load(f)
+    models = sorted(
+        models,
+        key=lambda d: (
+            d["Upload Time"],
+            d["Model ID"],
+            d["Model Version"],
```
dhrubo-os, line 158: can we keep `model version` before `model id`?

.github/workflows/model_uploader.yml
thanawan-atc: The workflow will exit with a failed status.

utils/model_uploader/model_autotracing.py
thanawan-atc: Sure! I can add another check. If any of the code throws the exception, the workflow will exit with failed status.

utils/model_uploader/update_models_upload_history_md.py
thanawan-atc: Yes!

.github/workflows/model_uploader.yml
dhrubo-os: Then we can start another workflow for the same model as a new independent workflow?

thanawan-atc: Do you mean after denial? If so, we can.

Issue #218, opened 2023-08-10
[FEATURE] Request for pretrained E5-small-v2 model


tallytarik:
**Is your feature request related to a problem?**
I would like to be able to use the [intfloat/e5-small-v2](https://huggingface.co/intfloat/e5-small-v2) embedding model with OpenSearch.

**What solution would you like?**
The OpenSearch Project can trace these models and upload them to its [artifacts server](https://opensearch-project.github.io/opensearch-py-ml/reference/pre_trained_models.html).

**What alternatives have you considered?**
Tracing the models myself and self-hosting them.

**Do you have any additional context?**
See https://github.com/opensearch-project/ml-commons/issues/1145 and https://github.com/opensearch-project/opensearch-py-ml/issues/182

PR #219, opened 2023-08-11
Handle undefined model max length

thanawan-atc:
### Description
For some model, `model.tokenizer.model_max_length` is not properly defined (i.e., set to be a very large number). This led to truncation problems and caused model deployment to fail. 
- For more context, see [this huggingface/transformer issue](https://github.com/huggingface/transformers/issues/14561).
- For some models, the problem has already been fixed directly to the model repo (e.g. [intfloat/multilingual-e5-small](https://huggingface.co/intfloat/multilingual-e5-small/commit/960310a7742a4e0a250dd720466bed84fee8caac)), but some still have that problem (e.g. [intfloat/e5-small-v2](https://huggingface.co/intfloat/e5-small-v2/blob/main/tokenizer_config.json), [gpt](https://github.com/huggingface/transformers/issues/14561)]
 
Screenshot 2023-08-10 at 9 26 51 PM
Screenshot 2023-08-10 at 9 27 03 PM

### Issues Resolved
Helped in tracing `intfloat/e5-small-v2` in #218 
 
### Check List


opensearch_py_ml/ml_models/sentencetransformermodel.py
```diff
@@ -791,6 +791,13 @@ def save_as_pt(
             zip_file_name = str(model_id.split("/")[-1] + ".zip")
         zip_file_path = os.path.join(model_output_path, zip_file_name)
 
+        # handle undefined model_max_length in model's tokenizer (e.g. "intfloat/e5-small-v2" )
+        if model.tokenizer.model_max_length == 1000000000000000019884624838656:
```
dhrubo-os, line 795: Let's assign a static variable for this magic number and also add more comments including the github issue why we need this. 

In addition, I assume this large is same for every cases, but I was wondering if we should do something like:

model.tokenizer.model_max_length = min (model.tokenizer.model_max_length, model.get_max_seq_length())

Can you please check if that's applicable for every cases?

thanawan-atc: I just made some changes. I use `if model.tokenizer.model_max_lengt > model.get_max_seq_length()` instead of `min` function, so that we can add `print()` to let user know what's happening.

dhrubo-os:
Thanks for raising this PR.

rbhavna (approved):

PR #265, opened 2023-08-30
Enable model listing workflow to exclude some models 

thanawan-atc:
### Description
Enable model listing workflow to exclude old 1.0.0 models from the generated model listing
 
### Check List
  - [x] All tests pass


utils/model_uploader/model_listing/excluded_models.txt
```diff
@@ -0,0 +1,18 @@
+ml-models/huggingface/sentence-transformers/all-MiniLM-L12-v2/1.0.0/onnx
```
dhrubo-os, line 1: Can we keep these values in a python file with a constant value? May be we can reuse the same constant for the testing also. And in the python file I would like to have a detailed comment about this exclusion.

dhrubo-os:
conflict in the changelog.

thanawan-atc:
Migrated excluded_models to py file and already tested it with model listing workflow on my own environment

dhrubo-os (approved):

ylwu-amzn (approved):

PR #301, opened 2023-09-22
Fix issue with polling logic for registering models

### LINKED ISSUE #300: [BUG] Registering Model Waits Longer Than Necessary
**What is the bug?**
When registering a pretrained model, the status of a task is polled. Polling should terminate once the state is terminal, but continues until the timeout is reached.

**How can one reproduce the bug?**
- Enable INFO logging, so that you see REST requests being executed.
- Call the method `register_pretrained_model` in `ml_commons_client` with some input.
- Wait for log messages such as `opensearch:base.py:236 GET http://localhost:9200/_plugins/_ml/tasks/LTP4t4oBYTCTt2_NvEZS [status:200 request:0.005s]`
- Look at the status returned from that URL. After a while it will change from `CREATED` to `COMPLETED`
- The polling should stop now, but it continues until the timeout is reached

**What is the expected behavior?**
Once the task is `COMPLETED`, polling should stop.

**What is your host/environment?**
 - OS: Linux
 - Version: 1.1.0

**Do you have any screenshots?**
No.

**Do you have any additional context?**
I believe that that the root cause is an incorrect Boolean condition: There is polling logic in the lines https://github.com/opensearch-project/opensearch-py-ml/blob/66325f30e44fba2fb0d7d6141f832945eb773c19/opensearch_py_ml/ml_commons/ml_commons_client.py#L267 and https://github.com/opensearch-project/opensearch-py-ml/blob/66325f30e44fba2fb0d7d6141f832945eb773c19/opensearch_py_ml/ml_commons/ml_commons_client.py#L378.

Here `not task_flag` means "operation has not completed" and `time.time() < end` means "timeout has not expired".
Polling should only continue when *both* conditions are True. But as these are linked with `or`, polling continues also when only one of these conditions is True. So when the operation is complete, but the timeout has not expired yet, polling is continued.

The `or` should be changed to `and` in both lines.

upwards-gravity:
### Description
Fixes an issue with the polling logic for registering models that is described in #300.
 
### Issues Resolved
Fixes #300.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/opensearch_py_ml/ml_commons/ml_commons_client.py
+++ b/opensearch_py_ml/ml_commons/ml_commons_client.py
@@ -264,7 +264,7 @@ class MLCommonClient:
         )
         end = time.time() + TIMEOUT  # timeout seconds
         task_flag = False
-        while not task_flag or time.time() < end:
+        while not task_flag and time.time() < end:
             time.sleep(1)
             status = self._get_task_info(output["task_id"])
             if status["state"] != "CREATED":
@@ -375,7 +375,7 @@ class MLCommonClient:
         if wait_until_task_done:
             end = time.time() + TIMEOUT  # timeout seconds
             task_flag = False
-            while not task_flag or time.time() < end:
+            while not task_flag and time.time() < end:
                 time.sleep(1)
                 output = self._get_task_info(task_id)
                 if (


upwards-gravity:
Something seems to be wrong with the sign-off. I don't know what that is though, since I added `Signed-off-by: Tim Vossen <tim.vossen@intrafind.com>` to the only commit. :shrug:

opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -264,7 +264,7 @@ def _send_model_info(self, model_meta_json: dict):
         )
         end = time.time() + TIMEOUT  # timeout seconds
         task_flag = False
-        while not task_flag or time.time() < end:
+        while not task_flag and time.time() < end:
```
dhrubo-os, line 267: I believe we should consider the current condition. If the task status is anything other than 'created,' there's no need for us to continuously check the task status. If this condition is met before the set time limit, we can proceed further; otherwise, we will wait until the specified time has passed.

dhrubo-os:
> Something seems to be wrong with the sign-off. I don't know what that is though, since I added `Signed-off-by: Tim Vossen <tim.vossen@intrafind.com>` to the only commit. 🤷

Did you try to do: `git commit -m "your message" -s`

upwards-gravity:
> Did you try to do: `git commit -m "your message" -s`

I redid the commit like that and now it seems to work. Don't really see why, as the commit messages is identical, but whatever. 

Thanks!

upwards-gravity:
> I believe we should consider the current condition. If the task status is anything other than 'created,' there's no need for us to continuously check the task status. If this condition is met before the set time limit, we can proceed further; otherwise, we will wait until the specified time has passed.

I found this comment in my mails (but nowhere else) and I am not sure I understand. Was this meant as a confirmation of the changes I did? Or do you suggest to do something else, @dhrubo-os?

dhrubo-os:
> > I believe we should consider the current condition. If the task status is anything other than 'created,' there's no need for us to continuously check the task status. If this condition is met before the set time limit, we can proceed further; otherwise, we will wait until the specified time has passed.
> 
> I found this comment in my mails (but nowhere else) and I am not sure I understand. Was this meant as a confirmation of the changes I did? Or do you suggest to do something else, @dhrubo-os?

Please ignore that.

dhrubo-os (approved):

b4sjoo (approved):

PR #346, opened 2023-11-24
Add GitHub action for opensearch-py-ml release

zelinh:
### Description
Enable GHA release by adding trusted publisher.
This workflow will be triggered when a tag is pushed to this repository. The workflow will built the product and publish to Pypi.
The GitHub release will be created after publishing to Pypi with `artifacts.tar.gz`.

I will add this workflow as trusted publisher according to this. https://docs.pypi.org/trusted-publishers/adding-a-publisher/ 

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build-libraries/issues/222
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -9,6 +9,9 @@ on:
 jobs:
   draft-a-release:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v3
@@ -34,10 +37,11 @@ jobs:
         run: |
           python -m build
           tar -zvcf artifacts.tar.gz dist
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
       - name: Draft a release
         uses: softprops/action-gh-release@v1
         with:
-          draft: true
           generate_release_notes: true
           files: |
             artifacts.tar.gz


.github/workflows/release-drafter.yml
```diff
@@ -34,10 +37,11 @@ jobs:
         run: |
           python -m build
           tar -zvcf artifacts.tar.gz dist
+      - name: Publish package distributions to PyPI
```
dhrubo-os, line 40: Aren't we doing the same thing [here](https://github.com/opensearch-project/opensearch-py-ml/blob/c9779818ce3d1e16916891d182dcadcdd22f0f94/jenkins/release.JenkinsFile#L12)?

zelinh: Yes. The reason we did in Jenkins was that we signed our GPG signatures in Jenkins workflow. Since Pypi now support 2FA/trusted publisher and doesn't require GPG signatures anymore. There is no need for us to run Jenkins for release. 

Therefore, we could simply use GHA for release. https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-pypi

dhrubo-os: In that case don't we need to remove the jenkins workflow code? or it will trigger both workflow?

zelinh: Yes. We could remove the jenkins workflow or we can disable on Jenkins side as well. Maybe let's keep the jenkins code for now until we successfully release next time through GHA?

dhrubo-os: I'm planning to make a release in next month. So I just wanted to make sure if I try to release an artifact two workflow won't trigger together, that's all. If you can confirm, that won't be the case, I'm fine to approve this PR. Thanks.

zelinh: Yes, I will disable the jenkins workflow to avoid it triggered. 
For this GHA release workflow, I will need to add trusted publisher within our Pypi account to make sure it will run. 

I will make sure complete two of above tasks so that only GHA workflow will run for release once this is merged. Thanks.

dhrubo-os (approved):

rbhavna (approved):

zelinh:
Pypi trusted publisher is added.
Jenkins workflow https://build.ci.opensearch.org/job/opensearch-py-ml-release/ is disabled.

PR #361, opened 2023-12-19
Bug fix for: Only require model_id once to save a modelreq 

### LINKED ISSUE #323: Only require model_id once to save a model
In order to save an embedding model (which I will then upload to my OS cluster) I need to do the following steps:
```
model_name = Path(model_id).name
model = SentenceTransformerModel(
    model_id=model_id,
    folder_path=f"/tmp/{model_name}"
)
model.save_as_pt(
    sentences=["dummy sentence"],
    model_id=model_id,
    save_json_folder_path=f"/tmp/{model_name}"
)
mconfig_path = model.make_model_config_json(model.model_id, )
ml_client.register_model(
    model_path=f"/tmp/{model_name}/{model_name}.zip",
    model_config_path=mconfig_path
)
```
If I don't specify the model id (huggingface id) in both places I end up with `"sentence-transformers/msmarco-distilbert-base-tas-b"`, presumably because that's what they used in that [blog post](https://opensearch.org/blog/semantic-science-benchmarks/). This can be smoother.

**What solution would you like?**
I would like to EITHER 1/ only provide the model_id when I instantiate the SentenceTransformerModel and it gets used in all class methods where applicable, rather than randomly defaulting back to "distilbert-tas-b" OR 2/ use `save_as_pt` as a static method and never instantiate the class at all. Also it would be nice if I could specify the output location of `make_model_config_json`.

**What alternatives have you considered?**

**Do you have any additional context?**

Vinay-Vinod:
### Description
In this pull request, I implemented a bug fix in opensearch_py_ml/ml_models/sentencetransformermodel.py to ensure that model_id is only required once for saving a model. Comprehensive tests were added in tests/ml_commons/test_ml_commons_client.py and tests/ml_models/test_sentencetransformermodel_pytest.py to validate the functionality.
 
### Issues Resolved
#323 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/tests/ml_models/test_sentencetransformermodel_pytest.py
+++ b/tests/ml_models/test_sentencetransformermodel_pytest.py
@@ -8,11 +8,12 @@
 import json
 import os
 import shutil
+from unittest.mock import MagicMock, patch
 from zipfile import ZipFile
 
 import pytest
 
-from opensearch_py_ml.ml_models import SentenceTransformerModel
+from opensearch_py_ml.ml_models.sentencetransformermodel import SentenceTransformerModel
 
 TEST_FOLDER = os.path.join(
     os.path.dirname(os.path.abspath("__file__")), "tests", "test_model_files"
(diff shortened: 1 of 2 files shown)


tests/ml_models/test_sentencetransformermodel_pytest.py
```diff
@@ -658,5 +660,23 @@ def test_zip_model_with_license():
     clean_test_folder(TEST_FOLDER)
 
 
+@pytest.fixture
+def mock_opensearch_client():
```
rawwar, line 664: We might not want to mock Opensearch client. Tests we are running are integration tests. Hence, mocking might not be useful.

rawwar: Seems like I am wrong. In this case, we don't need the client.

dhrubo-os: @Vinay-Vinod Thanks for raising the PR. Could you please make sure if this PR addresses all these:

```
What solution would you like?
I would like to EITHER 1/ only provide the model_id when I instantiate the SentenceTransformerModel and it gets used in all class methods where applicable, rather than randomly defaulting back to "distilbert-tas-b" OR 2/ use save_as_pt as a static method and never instantiate the class at all. Also it would be nice if I could specify the output location of make_model_config_json.
```

1. only provide the model_id when I instantiate the SentenceTransformerModel and it gets used in all class methods where applicable, rather than randomly defaulting back to "distilbert-tas-b"
2. In `save_as_pt` method, if no other model id is given as function parameter, it will use the model id which was initialized with SentenceTransformerModel class. If still a model is given as a parameter then in  `save_as_pt` method we'll use that model id. Please let me know if you have any confusion.
3. it would be nice if I could specify the output location of make_model_config_json  --> Can we take care of this one too? Or are you planning to raise another PR for this?

PR #375, opened 2024-02-13
add cross-encoder tracing, config-generating, and uploading

HenryL27:
### Description
Adds capabilities for uploading cross encoders. Introduces a 'CrossEncoderModel' class with methods to zip (either torchscript or onnx) and generate a json configuration (as in SentenceTransformerModel) along with a utility `upload` function. Example
```python
from opensearch_py_ml.ml_models import CrossEncoderModel
from opensearchpy import OpenSearch

client = OpenSearch(**os_client_args)
model = CrossEncoderModel("BAAI/bge-reranker-base", overwrite=True)
model.upload(client, framework='pt')
model.upload(client, framework='onnx')
```
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


dhrubo-os:
Thanks for raising this PR. We need to add corresponding tests too. Seems like few functions are mostly common to the `SentenceTransformerModel` class. May be we can take out the common functionality to some other utils class?

dhrubo-os:
Customer should have a way to give the name of the generated zip file and the model file as like we have in our SentenceTransformerModel class

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,308 @@
...
+            model_loc = CrossEncoderModel._trace_onnx(model, features, mname)
+        else:
+            raise Exception(
+                f"Unrecognized framework {framework}. Accepted values are `pt`, `onnx`"
+            )
+
+        # save tokenizer file
+        tk_path = Path(f"/tmp/{mname}-tokenizer")
+        tk.save_pretrained(tk_path)
+        _fix_tokenizer(tk.model_max_length, tk_path)
```
dhrubo-os, line 123: `tk.model_max_length` --> not sure if we will get this value all the time. Please take a look at this PR: https://github.com/opensearch-project/opensearch-py-ml/pull/219

dhrubo-os:
Tests are failing.

opensearch_py_ml/ml_models/crossencodermodel.py
HenryL27: k. going with the strategy proposed in https://github.com/huggingface/transformers/issues/14561 - look for `max_position_embeddings` or `n_positions` in the config object and if I miss in both those places, I set to 32k and hope it's fine. (The fix using sentence_transformers `model.get_max_seq_length()` also has the potential to return None)

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,319 @@
...
+        model_config_content = {
+            "name": model_name,
+            "version": f"1.0.{version_number}",
+            "description": description,
+            "model_format": model_format,
+            "function_name": "TEXT_SIMILARITY",
+            "model_content_hash_value": hash_value,
+            "model_config": {
+                "model_type": model_type,
+                "embedding_dimension": 1,
```
dhrubo-os, line 275: Is this correct?

dhrubo-os: If you look at here: https://huggingface.co/BAAI/bge-reranker-base/blob/main/config.json

max_position_embeddings = 514

HenryL27: yes, this is what I did. Artifact of the implementation (depends heavily on the embedding model code). Can probably be cleaned up a bit

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,319 @@
...
+        tk_path = Path(f"/tmp/{mname}-tokenizer")
+        tk.save_pretrained(tk_path)
+        if tk.model_max_length > model.get_max_length():
+            model_config = AutoConfig.from_pretrained(self._hf_model_id)
+            if hasattr(model_config, "max_position_embeddings"):
+                tk.model_max_length = model_config.max_position_embeddings
+            elif hasattr(model_config, "n_positions"):
+                tk.model_max_length = model_config.n_positions
+            else:
+                tk.model_max_length = 2**15  # =32768. Set to something big I guess
```
dhrubo-os, line 130: Setting an arbitrary value doesn't seem like a good solution.

What do you think about following this: https://github.com/opensearch-project/opensearch-py-ml/blob/main/opensearch_py_ml/ml_models/sentencetransformermodel.py#L936-L942

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,319 @@
...
+        return save_loc
+
+    @staticmethod
+    def _trace_onnx(model, features, mname):
+        """
+        Compiles the model to ONNX format.
+        """
+        # export to onnx
+        save_loc = Path(f"/tmp/{mname}.onnx")
+        torch.onnx.export(
```
dhrubo-os, line 180: i think we should add `onnx` in the requirements.txt. 

In addition, can we also add in the `requirements-dev.txt`?

opensearch_py_ml/ml_models/crossencodermodel.py
HenryL27: would love to. Unfortunately, `model.get_max_seq_length()` is not a method of most hf transformer ModelForSequenceClassification - that's a special thing from the sentence-transformers model interface, and I'm not sure it is the guarantee that it claims to be: [[implementation]](https://github.com/UKPLab/sentence-transformers/blob/1eec036cdcce649502de6152cb8a3b1ea1bb9a8c/sentence_transformers/SentenceTransformer.py#L436)

opensearch_py_ml/ml_models/crossencodermodel.py
HenryL27: looks like it's there already?

opensearch_py_ml/ml_models/crossencodermodel.py
dhrubo-os: I'm seeing two problems here:

1. tk.model_max_length is None:    https://github.com/huggingface/transformers/issues/14561 ; if you see this, `model_max_length` will be a larger value and we are going to set a large value which model can't process

2. I'm not a big fan of setting a static big value like `tk.model_max_length = 2**15`. What exactly are we achieving here?

It's important to align `model_max_length` with the model's actual capability (`max_position_embeddings`). Setting `model_max_length` to a higher value than `max_position_embeddings` could lead to errors or unexpected behavior, as the model won't be able to handle inputs longer than its architectural limit. Conversely, setting a lower `model_max_length` can be useful for optimizing performance or adhering to specific task requirements.

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,319 @@
...
+        tk = AutoTokenizer.from_pretrained(self._hf_model_id)
+        model = AutoModelForSequenceClassification.from_pretrained(self._hf_model_id)
+        features = tk([["dummy sentence 1", "dummy sentence 2"]], return_tensors="pt")
+        mname = Path(self._hf_model_id).name
+
+        # bge models don't generate token type ids
+        if mname.startswith("bge"):
+            features["token_type_ids"] = torch.zeros_like(features["input_ids"])
+
+        if framework == "pt":
```
dhrubo-os, line 109: i think we should accept "torch_script" as a framework parameter instead of `pt`. `pt` is an extension type but not a framework. We decide to convert as .pt, somebody else might want to decide as `pth` or `ptc`. Customer might not know the file format of the torch_script. 

So framework choice should be: `torch_script` or `onnx`

HenryL27: yep

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,323 @@
...
+        if model_format is None:
+            raise Exception(
+                "Model format either not found or not supported. Zip the model before generating the config"
+            )
+        model_config_content = {
+            "name": model_name,
+            "version": f"1.0.{version_number}",
+            "description": description,
+            "model_format": model_format,
+            "function_name": "TEXT_SIMILARITY",
```
dhrubo-os, line 275: we need `model_task_type`. Let's follow sentencetransformer model example method.

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,323 @@
...
+            model_format = {"torch_script": "TORCH_SCRIPT", "onnx": "ONNX"}.get(
+                self._framework
+            )
+        if model_format is None:
+            raise Exception(
+                "Model format either not found or not supported. Zip the model before generating the config"
+            )
+        model_config_content = {
+            "name": model_name,
+            "version": f"1.0.{version_number}",
```
dhrubo-os, line 272: we don't need to append `1.0`.

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,324 @@
...
+            raise Exception(
+                "No model zip file. Generate the model zip file before generating the config."
+            )
+        if not self._model_zip.exists():
+            raise Exception(f"Model zip file {self._model_zip} could not be found")
+        hash_value = _generate_model_content_hash_value(str(self._model_zip))
+        if model_name is None:
+            model_name = Path(self._hf_model_id).name
+        if description is None:
+            description = f"Cross Encoder Model {model_name}"
```
dhrubo-os, line 255: Can we follow the way we are showing description for embedding models?

opensearch_py_ml/ml_models/crossencodermodel.py
HenryL27: I don't understand these issues. `tk.model_max_length is None` is exactly the condition that triggers setting it to something reasonable. I agree that it's important to set it to a value that agrees with the model's capability: that's why the first condition here is checking and setting to `max_position_embeddings`. We only pick the very large number when we don't have a bound. For instance, suppose someone trained a mamba-based cross encoder with infinite context cuz it's a RNN.

dhrubo-os:
Looks like linting and integration tests are failing could you please take a look?

asfoorial:
This PR fails with this promising rerank model https://huggingface.co/mixedbread-ai/mxbai-rerank-xsmall-v1

I tried both torch_script and onnx. The torch_script failed to continue saving the trace giving
Could not export Python function call ‘XSoftmax’. Remove calls to Python functions before export. Did you forget to add @script or @script_method annotation? If this is a nn.ModuleList, add it to constants:

While the onnx saved the onnx and zip files but failed during prediction giving
{
“error”: {
“root_cause”: [
{
“type”: “translate_exception”,
“reason”: “translate_exception: java.lang.IllegalArgumentException: Input mismatch, looking for: [input_ids, attention_mask]”
}
],
“type”: “m_l_exception”,
“reason”: “m_l_exception: Failed to inference TEXT_SIMILARITY model: ZwQzA44BK0CtZG_zHaFw”,
“caused_by”: {
“type”: “privileged_action_exception”,
“reason”: “privileged_action_exception: null”,
“caused_by”: {
“type”: “translate_exception”,
“reason”: “translate_exception: java.lang.IllegalArgumentException: Input mismatch, looking for: [input_ids, attention_mask]”,
“caused_by”: {
“type”: “illegal_argument_exception”,
“reason”: “Input mismatch, looking for: [input_ids, attention_mask]”
}
}
}
},
“status”: 500
}

HenryL27:
looks like maybe https://github.com/huggingface/transformers/issues/20815 is related? Is mxbai-rerank based on Deberta do you know? If so try setting `model_type='deberta'` in the make_model_config_json fn call

asfoorial:
> looks like maybe [huggingface/transformers#20815](https://github.com/huggingface/transformers/issues/20815) is related? Is mxbai-rerank based on Deberta do you know? If so try setting `model_type='deberta'` in the make_model_config_json fn call

Yes it is based on deberta.

asfoorial:
> looks like maybe [huggingface/transformers#20815](https://github.com/huggingface/transformers/issues/20815) is related? Is mxbai-rerank based on Deberta do you know? If so try setting `model_type='deberta'` in the make_model_config_json fn call

I have set model type to deberta and converted the model into onnx (torch script did not work) and was able to deploy it. However I get the same error above alwhen calling _predict API

opensearch_py_ml/ml_models/crossencodermodel.py
```diff
@@ -0,0 +1,406 @@
...
+        :type zip_fname: str
+        :return: the path with the zipped model
+        :rtype: Path
+        """
+        tk = AutoTokenizer.from_pretrained(self._hf_model_id)
+        model = AutoModelForSequenceClassification.from_pretrained(self._hf_model_id)
+        features = tk([["dummy sentence 1", "dummy sentence 2"]], return_tensors="pt")
+        mname = Path(self._hf_model_id).name
+
+        # bge models don't generate token type ids
```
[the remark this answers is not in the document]
HenryL27: I arrived at this conclusion by trying to do the thing and failing, so there might be an issue somewhere out there but it's more of a fundamental architectural feature, not a bug

opensearch_py_ml/ml_models/crossencodermodel.py
HenryL27: Interesting. Can we just let huggingface/transformers fix this bug? It seems like a them problem, and from what I can tell the only times we're gonna hit it is if someone is trying to use a very old tokenizer file with their thing. At that point I hope we can assume the user is proficient enough with transformers to debug if necessary.

Issue #287, opened 2023-09-17
[FEATURE] Add ML-Commons Predict api functionality in py-ml


dhrubo-os:
Method name will be predict()

Parameters: [predict request](https://opensearch.org/docs/latest/ml-commons-plugin/api/#request-1)

    algorithm name
    model id
    predict object (json)

Response needs to be similar [like this](https://opensearch.org/docs/latest/ml-commons-plugin/api/#response-4)

Depending on the algorithm, response can be different. 

We can write this function [here](https://github.com/opensearch-project/opensearch-py-ml/blob/main/opensearch_py_ml/ml_commons/ml_commons_client.py)

And need to add unit test [here](https://github.com/opensearch-project/opensearch-py-ml/tree/main/tests/ml_commons)

plzdoo:
I would like to contribute to this issue

plzdoo:
I would like to contribute to this issue

dhrubo-os:
Sure, assigned to you.

PR #425, opened 2024-11-06
Implement `predict` method and add unit tests

Yerzhaisang:
### Description
- Implemented the `predict` method in `MLCommonClient`.
- Added unit tests to ensure the correctness of the `predict` method's behavior.
- Integrated the method with the existing system and ensured it handles the parameters (`text_docs`, `return_number`, `target_response`) as expected.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-py-ml/issues/287
 
### Check List
  - [x] All tests pass

### DIFF
--- a/opensearch_py_ml/ml_commons/ml_commons_client.py
+++ b/opensearch_py_ml/ml_commons/ml_commons_client.py
@@ -567,6 +567,36 @@ class MLCommonClient:
             url=API_URL,
             body=API_BODY,
         )
+    
+    def predict(
+        self, 
+        algorithm_name: str, 
+        model_id: str, 
+        predict_object: dict
+    ) -> dict:
+        """
+        Generalized predict method to make predictions using different ML algorithms.
+
+        :param algorithm_name: The name of the algorithm, e.g., 'kmeans', 'text_embedding'
+        :type algorithm_name: str
+        :param model_id: Unique identifier of the deployed model
+        :type model_id: str
+        :param predict_object: JSON object containing the input data and parameters for prediction
+        :type predict_object: dict
+        :return: Prediction response from the ML model
+        :rtype: dict
+        """
+        # Construct the URL for the prediction request
+        url = f"{ML_BASE_URI}/_predict/{algorithm_name}/{model_id}"
+        
+        # Make the POST request to the prediction API
+        response = self._client.transport.perform_request(
+            method="POST",
+            url=url,
+            body=predict_object
+        )
+
+        return response
 
     def undeploy_model(self, model_id: str, node_ids: List[str] = []) -> object:
         """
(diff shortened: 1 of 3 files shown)


opensearch_py_ml/ml_commons/ml_commons_client.py
```diff
@@ -568,6 +568,29 @@ def unload_model(self, model_id: str, node_ids: List[str] = []) -> object:
             body=API_BODY,
         )
 
+    def predict(self, algorithm_name: str, model_id: str, predict_object: dict) -> dict:
```
dhrubo-os, line 571: We have a method [generate_model_inference](https://github.com/opensearch-project/opensearch-py-ml/blob/39284d6e99cd9b449bcfe99511c68a63d897ca78/opensearch_py_ml/ml_commons/ml_commons_client.py#L501-L517), which is in fact the predict functionality. We can just invoke this method here? 

We could change the method name `generate_model_inference` to predict, but for backward compatibility keep the method as it is.

Yerzhaisang: `generate_model_inference ` and `predict` methods are not the same because:
1) URLs are not the same
2) `generate_model_inference ` doesn't use `algorithm_name`

Yerzhaisang: could you please take a look at [the issue description](https://github.com/opensearch-project/opensearch-py-ml/issues/287)?

Yerzhaisang: I see @dhrubo-os, thank you for the feedback!

dhrubo-os:
codecov/patch — 33.33% of diff hit (target 91.53%) 
 
 @Yerzhaisang could you please take a look?

Yerzhaisang:
@dhrubo-os it's ready for overview

opensearch_py_ml/ml_commons/ml_commons_client.py
b4sjoo: So aren't we going to remove the `algorithm_name` here?

dhrubo-os: Good point. I think it's better to keep this as a optional parameter for the BWC as before we needed to have algorithm_name for model prediction.


@Yerzhaisang currently algorithm_name isn't optional here. Can we do something like `def predict(self, model_id: str, predict_object: dict, algorithm_name: str = None) -> dict:` ?

Yerzhaisang: done

dhrubo-os (approved):

b4sjoo (approved):

PR #475, opened 2025-04-18
Fix integ test flaky test

zhichao-aws:
### Description
We see integ test failing https://github.com/opensearch-project/opensearch-py-ml/actions/runs/14520924174/job/40742876990, and the error message is 
```
E           huggingface_hub.errors.HfHubHTTPError: 429 Client Error: Too Many Requests for url: https://huggingface.co/opensearch-project/opensearch-neural-sparse-encoding-doc-v3-distill/resolve/main/config.json
```

The error is caused by huggingface throttling. This PR makes all sparse model tests use one model id, therefore we can reuse the model file in cache, to reduce the probability of flaky test.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/tests/ml_models/test_sparseencondingmodel_pytest.py
+++ b/tests/ml_models/test_sparseencondingmodel_pytest.py
@@ -278,7 +278,7 @@ def test_save_as_pt_with_license():
 
 
 def test_default_description():
-    model_id = "opensearch-project/opensearch-neural-sparse-encoding-doc-v3-distill"
+    model_id = "opensearch-project/opensearch-neural-sparse-encoding-doc-v2-distill"
     model_format = "TORCH_SCRIPT"
     expected_model_description = "This is a neural sparse encoding model: It transfers text into sparse vector, and then extract nonzero index and value to entry and weights. It serves only in ingestion and customer should use tokenizer model in query."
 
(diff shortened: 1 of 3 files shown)


tests/ml_models/test_sparseencondingmodel_pytest.py
```diff
@@ -25,6 +27,35 @@
 )
 
 
+def retry(n_retries=3, delay_sec=10, exceptions=(Exception,)):
```
dhrubo-os, line 30: Will that increase the total integration test run time?

dhrubo-os:
Why do we need to download the model every time?

tests/ml_models/test_sparseencondingmodel_pytest.py
zhichao-aws: If we success to download the model at first time, it won't have impact on test time.

But if the first time get throttled, then yes, we'll wait extra times for retry.

zhichao-aws:
> Why do we need to download the model every time?

Because every time we trigger integ test CI, it starts a new instance and run all test cases. On this instance, it doesn't have local file cache of these models, so it need to download the model file when first called to init the model.

But for this instance, it only download the model once. The following init code just read from local cache.

dhrubo-os (approved):

b4sjoo (approved):

PR #486, opened 2025-04-28
Add semantic highlighter model auto tracing support

junqiu-lei:
### Description
Add support to auto tracing semantic highlighter model.

### Test
Tested CI at https://github.com/junqiu-lei/opensearch-py-ml/actions/runs/14740372844/job/41376974972
 
### Issues Resolved
Part of https://github.com/opensearch-project/neural-search/issues/1182
 
### Check List
  - [x] All tests pass
  - [x] New functionality has javadoc added


.ci/run-opensearch.sh
```diff
@@ -97,7 +97,7 @@ END
     docker run \
       --name "$node_name" \
       --network "$network_name" \
-      --env "ES_JAVA_OPTS=-Xms1g -Xmx1g" \
+      --env "OPENSEARCH_JAVA_OPTS=-Xms4g -Xmx4g" \
```
zhichao-aws, line 100: What's the instance we use for the CI? Does the instance has enough memory for 4g heap size?

.github/workflows/integration.yml
```diff
@@ -14,10 +14,10 @@ jobs:
       fail-fast: false
       matrix:
         cluster: ["opensearch"]
-        secured: ["true"]
+        secured: ["false"]
         entry:
           - { opensearch_version: 2.7.0 }
-          - { opensearch_version: 2.11.0 }
+          - { opensearch_version: 3.0.0-beta1 }
```
zhichao-aws, line 20: Can we add `3.0.0-beta1` and keep `2.11.0`, too? We want to make sure the model is bwc on previous version (except the new introduced types)

utils/model_uploader/semantic_highlighter_autotracing.py
```diff
@@ -0,0 +1,345 @@
...
+    trace_dataset = Dataset.from_dict(
+        {
+            "question": [[question]],
+            "context": [context],
+            "word_level_sentence_ids": [sentence_ids],
+            "orig_sentence_labels": [orig_sentence_labels],
+            "id": ["test"],
+        }
+    )
+    base_model_id = "bert-base-uncased"
```
zhichao-aws, line 122: we shouldn't hard code the base_model_id, it should be consistent with the MODEL_ID in CI

utils/model_uploader/semantic_highlighter_autotracing.py
```diff
@@ -0,0 +1,345 @@
...
+    # jointly tokenize questions and context
+    tokenized_examples = tokenizer(
+        examples["question"],
+        examples["context"],
+        truncation="only_second",
+        max_length=max_seq_length,
+        stride=stride,
+        return_overflowing_tokens=True,
+        padding=padding,
+        is_split_into_words=True,
```
zhichao-aws, line 52: Here we set the is_split_into_words=True, but in L115 the input question is a sentence?

utils/model_uploader/semantic_highlighter_autotracing.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+def prepare_train_features(
+    tokenizer, examples, max_seq_length=512, stride=128, padding=False
+):
+    # jointly tokenize questions and context
+    tokenized_examples = tokenizer(
+        examples["question"],
+        examples["context"],
+        truncation="only_second",
```
zhichao-aws, line 47: what if the first sentence is longer than 512 tokens? Is it able to work without context?

opensearch_py_ml/ml_models/semantic_highlighter_model.py
```diff
@@ -0,0 +1,357 @@
...
+
+from opensearch_py_ml.ml_commons.ml_common_utils import (
+    _generate_model_content_hash_value,
+)
+from opensearch_py_ml.ml_models.base_models import BaseUploadModel
+
+DEFAULT_MODEL_ID = "opensearch-project/opensearch-semantic-highlighter-v1"
+
+
+class TraceableBertTaggerForSentenceExtractionWithBackoff(BertPreTrainedModel):
```
zhichao-aws, line 25: I see we use `BertTaggerForSentenceExtraction` at https://huggingface.co/opensearch-project/opensearch-semantic-highlighter-v1.

Why don't we re-use the BertTaggerForSentenceExtraction class, instead of defining classifier and forward function here?

opensearch_py_ml/ml_models/semantic_highlighter_model.py
```diff
@@ -0,0 +1,357 @@
...
+    A torch.jit-compatible version of the sentence highlighter model.
+    For inference only. Classification logic includes backoff rule
+    with minimum confidence alpha=0.05.
+    """
+
+    def __init__(self, config):
+        super().__init__(config)
+        self.num_labels = config.num_labels
+
+        self.bert = BertModel(config).from_pretrained(
```
zhichao-aws, line 36: If we use from_pretrained to init the model, we don't need to init from `bert-base-uncased` at __init__ function

opensearch_py_ml/ml_models/semantic_highlighter_model.py
```diff
@@ -0,0 +1,357 @@
...
+        model = model.to(device)
+
+        # Save tokenizer files
+        tokenizer_path = os.path.join(self.folder_path, "tokenizer")
+        os.makedirs(tokenizer_path, exist_ok=True)
+        tokenizer.save_pretrained(tokenizer_path)
+        print(f"Tokenizer files saved to {tokenizer_path}")
+
+        # Move inputs to CPU
+        for k, v in example_inputs.items():
```
zhichao-aws, line 231: It's on CPU by default. Move tensors to GPU need we explicitly set it

zhichao-aws (commented):
Leave some comments. Have you tried to run it at the forked repo, with skip_deployment=False?

junqiu-lei:
> Leave some comments. Have you tried to run it at the forked repo, with skip_deployment=False?

Thanks @zhichao-aw. Yes, this is the test CI from my forked repo https://github.com/junqiu-lei/opensearch-py-ml/actions/runs/14740372844/job/41376974972

I also have sanity tested inference api call with above output model artifact.

.ci/run-opensearch.sh
junqiu-lei: Yes, the standard github runner we are using has 16 gb memory.

I checked on the CI https://github.com/junqiu-lei/opensearch-py-ml/actions/runs/14740372844/job/41376688106

Their doc: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

.github/workflows/integration.yml
junqiu-lei: Sure, will update.

utils/model_uploader/semantic_highlighter_autotracing.py
junqiu-lei: The base model here is not the target trace model(semantic highlighter model). Will add comment in code.

utils/model_uploader/semantic_highlighter_autotracing.py
junqiu-lei: The model is trained in words(token) level, when do inference, the input is in sentence level. cc: @greaa-aws

utils/model_uploader/semantic_highlighter_autotracing.py
greaa-aws: The tokenizer will throw an `Exception: Truncation error` if the first input exceeds the truncation length under an `only_second` truncation strategy. We do not expect the highlighter query to exceed 512 tokens in the vast majority of cases, but maybe we should call out this limitation somewhere.

utils/model_uploader/semantic_highlighter_autotracing.py
greaa-aws: This argument really just signals to the tokenizer that it should treat a list of strings as a single text input, rather than a batch of inputs to be tokenized. You can check that the result is the same as if we had also split the input question into words.

opensearch_py_ml/ml_models/semantic_highlighter_model.py
greaa-aws: Agree that this is redundant in the context of the script. Note however the function itself doesn't make any assumption about the device of `example_inputs`.

opensearch_py_ml/ml_models/semantic_highlighter_model.py
greaa-aws: There are two differences between these classes:
1. The class defined here is traceable with `torch.jit`, whereas `BertTaggerForSentenceExtraction` is not.
2. The class defined here implements a different classification rule for selecting relevant sentences, which results in improved recall on our benchmarks.

opensearch_py_ml/ml_models/semantic_highlighter_model.py
greaa-aws: Agree this isn't necessary for the tracing script.

junqiu-lei:
With above code change, tested new round CI at https://github.com/junqiu-lei/opensearch-py-ml/actions/runs/14765830582/job/41456952166 which also includes model inference call verification.

Thanks @dhrubo-os @zhichao-aws @greaa-aws

zhichao-aws (approved):

dhrubo-os:
@junqiu-lei test coverage is low. Can we please write some unit tests?

opensearch_py_ml/ml_models/semantic_highlighter_model.py
```diff
@@ -0,0 +1,350 @@
...
+                zipObj.write(file_path, arcname=file)
+
+        if add_apache_license:
+            super()._add_apache_license_to_model_zip_file(zip_file_path)
+
+        self.torch_script_zip_file_path = zip_file_path
+        print(f"Zip file saved to {zip_file_path}")
+        return zip_file_path
+
+    def save_as_onnx(
```
dhrubo-os, line 259: What is the bottleneck for onnx? Have we tried to export it onnx?

noxfile.py
```diff
@@ -183,3 +183,25 @@ def sparsetrace(session):
...
+@nox.session(python=["3.9"])
+def semantic_highlighter_trace(session):
+    session.install(
+        "-r",
+        "requirements-dev.txt",
+        "--timeout",
+        "1500",
+    )
+    # Download necessary NLTK data
+    session.log("Downloading NLTK data (punkt_tab)")
```
dhrubo-os, line 197: Can we please add comment why do we need this?

utils/model_uploader/autotracing_utils.py
```diff
@@ -313,9 +314,14 @@ def prepare_files_for_uploading(
...
-        shutil.rmtree(folder_to_delete)
+        if os.path.exists(folder_to_delete):
+            shutil.rmtree(folder_to_delete)
+            print(f"Successfully deleted temporary folder: {folder_to_delete}")
+        else:
+            print(f"Temporary folder not found, skipping deletion: {folder_to_delete}")
     except Exception as e:
-        assert False, f"Raised Exception while deleting {folder_to_delete}: {e}"
+        # Changed from AssertionError to a warning
+        print(f"Warning: Raised Exception while deleting {folder_to_delete}: {e}")
```
dhrubo-os, line 324: Why do we want this?

opensearch_py_ml/ml_models/semantic_highlighter_model.py
```diff
@@ -0,0 +1,350 @@
...
+            input_ids,
+            attention_mask=attention_mask,
+            token_type_ids=token_type_ids,
+        )
+
+        sequence_output = outputs[0]
+        sequence_output = self.dropout(sequence_output)
+
+        @torch.jit.script_if_tracing
+        def _get_agg_output(ids, sequence_output):
```
dhrubo-os, line 58: Can we please add more comments if possible? 

Same comment applies for `semantic_highlighter_autotracing.py`.

opensearch_py_ml/ml_models/semantic_highlighter_model.py
junqiu-lei: Support onnx format isn't in scope of p0 release, we can look into it later. cc: @greaa-aws

utils/model_uploader/autotracing_utils.py
junqiu-lei: Got issue when in dev. Updated, no need this change anymore.

opensearch_py_ml/ml_models/semantic_highlighter_model.py
junqiu-lei: Added docs in code.

junqiu-lei:
> @junqiu-lei test coverage is low. Can we please write some unit tests?

Added

dhrubo-os (approved):

greaa-aws (approved):

